PR #12, opened 2021-01-25
formatting etc - oliver

kw-fn:
mixture of line justifications for readability and a couple spelling adjustments

thanks!

### DIFF
--- a/src/socketio/transport.rs
+++ b/src/socketio/transport.rs
@@ -20,10 +20,10 @@ use super::event::Event;
 /// The type of a callback function.
 pub(crate) type Callback<I> = RwLock<Box<dyn Fn(I) + 'static + Sync + Send>>;
 
-/// Represents an Ack which is given back to the caller.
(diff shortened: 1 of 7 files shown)


src/socketio/transport.rs
```diff
@@ -383,7 +385,12 @@ mod test {
         };
 
         socket
-            .emit_with_ack("test".into(), "123", Duration::from_secs(10), ack_callback)
+            .emit_with_ack(
+                "test".into(),
+                "123",
+                Duration::from_secs(10),
+                ack_callback,
+            )
```
1c3t3a, line 393: Could you run `cargo fmt` another time? This gets linted, but it would be nice to keep the checks green ;)

1c3t3a (changes requested):
Great thank you! Always good to have a second pair of eyes looking at the comments. There is just one thing to change and then we are ready to merge it.

1c3t3a:
Alright, thanks!

PR #20, opened 2021-02-14
Refactored packet decoding

1c3t3a:
This PR refactors the way the `socket.io` packet decoder parses the packet. Till now this was done via parsing the packet from a string, which involved parsing an utf-8 string. This PR changes this by parsing the packet from pure bytes.


Colin1860 (approved):
LGTM

PR #25, opened 2021-03-08
Implemented binary events

### LINKED ISSUE #24: Add the `BinaryEvent` and the `BinaryAck` message type
Till now the binary messages are ignored. This should be changed.

1c3t3a:
This PR closes #24 by adding support for binary events. 

This PR therefore introduces the enum `Payload`, which could either hold a `Binary` payload or a `String` payload. Generally every event interaction with some sort of payload is now handled through this enum. This PR also adds support for the Event Types `BinaryAck` and `BinaryEvent` which could both be used to send binary payload with a normal event or ack.

Last but not least the error enum was changed to make use of the proc_macros defined by the library `thiserror`.


src/lib.rs
```diff
@@ -240,71 +269,81 @@ impl Socket {
...
-        data: &str,
+        event: E,
+        data: D,
         timeout: Duration,
         callback: F,
     ) -> Result<()>
     where
-        F: FnMut(String) + 'static + Send + Sync,
+        F: FnMut(Payload) + 'static + Send + Sync,
+        E: Into<Event>,
```
Colin1860, line 342: Uncle Bob says in Clean Code that no method needs more than 3 parameters. I don't know, but maybe some intermediate struct would make sense here? But then again, its just Uncle Bob Martin

Colin1860 (approved):
LGTM

src/lib.rs
1c3t3a: Mh, don't think that's an issue, when I look at some other socketio client APIs I don't see a different layout. As a user it would make more trouble to build up some intermediate struct that can't be used elsewhere.

PR #52, opened 2021-07-16
feat: refactor

nshaaban-cPacket:
Refactoring to allow for easier implementation of a socket.io server. Might be wise to merge into a non-main branch (like 2.0.0), as this refactor only adds value after the server side implementations are in place, and during the process of implementing a server may require additional breaking changes that make more sense to release at once. 

Wanted to touch base with you to make sure a refactoring effort is wanted, as well as the refactoring direction outlined below was the direction you wanted to go. If so I'll complete the relevant comments/tests associated with such PRs. I welcome any comments you might have!

Although I did change some functions names/locations, none of my edits change existing public methods in socketio. (Although there are some edits to public types in engineio).

@1c3t3a 

Structural changes:
- Adopt a similar code structure to that in the javascript reference implementation https://github.com/socketio/engine.io/tree/master/lib
  - Moved protocol behavior logic to socket from transport (as some of that is shared between client and server)
  - Split transports into their own types implementing a common trait (Transport) rather than using an enum
  - Created EventEmitter trait
- Moved all hard coded packets (such as during websocket probe) to creating a Packet object then calling encode (makes it easier to update code in case protocol numbers/format changes)
- Moved all client side only operations to a trait called client (so it is clear exactly which functions can be shared and which can't)
- Used alternate module importing for the engineio::transports module so only transports can access any individual transport directly

Cosmetic Changes (can be separated into a new PR rather easily):
- Replaced the engine_io_mode: String with root_path: bool to remove all references of socket.io from the engine.io module
- Added an implementation for Into<handshake> for packet
- renamed get_random_time to get_hashed_time
- !Removed restriction that open needs to be called on a polling transport 
- !Switched websockets to send binary during the probe

TODO before merge:
- [X] Add back tests
  - [ ] Fix tests
- [ ] Ensure existing moved comments still are accurate
- [ ] Add new comments
- [ ] Refactor socketio folder to match patterns outlined above (looks pretty good, transport should be reamed to socket at least)
- [ ] Should any of the engineio code be more visible than pub(crate)?


src/engineio/transports/websocket_secure.rs
```diff
@@ -0,0 +1,100 @@
...
+        writer.send_message(&message)?;
+
+        Ok(())
+    }
+
+    fn poll(&mut self, _: String) -> Result<Bytes> {
+        let mut receiver = self.client.lock().unwrap();
+
+        // if this is a binary payload, we mark it as a message
+        let received_df = receiver.recv_dataframe().unwrap();
```
1c3t3a, line 84: Those `Result`s could use the `?` operator.

src/engineio/transport_emitter.rs
```diff
@@ -0,0 +1,166 @@
...
+impl Transport for TransportEmitter {
+    fn emit(&self, address: String, data: Bytes, is_binary_att: bool) -> Result<()> {
+        self.transport.lock()?.emit(address, data, is_binary_att)
+    }
+
+    fn poll(&mut self, address: String) -> Result<Bytes> {
+        self.transport.lock()?.poll(address)
+    }
+
+    fn name(&self) -> Result<String> {
```
1c3t3a, line 154: This could probably return a `&'static str`.

src/engineio/transports.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn poll(&mut self, address: String) -> Result<Bytes> {
+        match &*self.transport_type.read()? {
+            TransportTypes::Websocket => self
+                .websocket_secure
+                .lock()?
+                .as_mut()
+                .unwrap()
+                .poll(address),
+            TransportTypes::WebsocketSecure => {
+                self.websocket.lock()?.as_mut().unwrap().poll(address)
```
1c3t3a, line 116: Did you maybe mix up `Websocket` and `WebsocketSecure` here?

src/engineio/transports.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    fn name(&self) -> Result<String> {
+        match &*self.transport_type.read()? {
+            TransportTypes::Websocket => self.websocket_secure.lock()?.as_ref().unwrap().name(),
+            TransportTypes::WebsocketSecure => self.websocket.lock()?.as_ref().unwrap().name(),
+            TransportTypes::Polling => self.polling.lock()?.as_ref().unwrap().name(),
+        }
+    }
```
1c3t3a, line 128: Do you think that it might be nicer to maybe have a `to_string` method on `TransportType`? As fat as I can see this would also make the `name` method on `Transport` obsolete.

src/engineio/transports/websocket.rs
```diff
@@ -0,0 +1,101 @@
...
+        sender.send_message(&message)?;
+
+        Ok(())
+    }
+
+    fn poll(&mut self, _: String) -> Result<Bytes> {
+        let mut receiver = self.receiver.lock().unwrap();
+
+        // if this is a binary payload, we mark it as a message
+        let received_df = receiver.recv_dataframe().unwrap();
```
1c3t3a, line 85: Same as with websocket_secure. The `?` operator could be used instead of `unwrap`.

src/engineio/socket.rs
```diff
@@ -1,62 +1,68 @@
...
 #[derive(Clone, Debug)]
 pub struct EngineSocket {
-    transport_client: Arc<RwLock<TransportClient>>,
+    pub(super) transport: Arc<Mutex<TransportEmitter>>,
+    pub connected: Arc<AtomicBool>,
+    last_ping: Arc<Mutex<Instant>>,
+    last_pong: Arc<Mutex<Instant>>,
+    host_address: Arc<Mutex<Option<String>>>,
+    connection_data: Arc<RwLock<Option<HandshakePacket>>>,
+    root_path: Arc<RwLock<Option<String>>>,
```
1c3t3a, line 30: Nice! That's way better than the old bool ;)

1c3t3a (commented):
Alright! That already looks decent. I wasn't able to look through every file as this makes quite a lot of changes, but I'm definitely going to within the next days. 

Concerning the refactoring in general: The structural changes using the traits are something I already planned on doing. I think the design is clever and straight forward while also being closer to the reference implementation. There is obviously some work to do, but this is a draft PR. I think restructuring the transport is a needed step before starting with the server implementation in order to be able to reuse as many code as possible. 

Concerning making engine.io public. I already planned to do this, but I was sure to refactor it beforehand. So thanks a lot! I would wait until the server is implemented in order to guarantee a stable API, though.

src/engineio/transports.rs
nshaaban-cPacket: Nice catch, fixed in https://github.com/1c3t3a/rust-socketio/pull/52/commits/b4a11b53a532fe8770c28308b4e79f2ae0d8d365

src/engineio/transports/websocket.rs
nshaaban-cPacket: Fixed https://github.com/1c3t3a/rust-socketio/pull/52/commits/87ccc34be6792e01ca25a28cb1ea5c0367e9b5b2

src/engineio/transports/websocket_secure.rs
nshaaban-cPacket: Fixed https://github.com/1c3t3a/rust-socketio/pull/52/commits/87ccc34be6792e01ca25a28cb1ea5c0367e9b5b2

src/engineio/transports.rs
nshaaban-cPacket: Hmmm, isn't to_string more of I want to see this object as a human readable string rather than seeing just it's type? Changed this function to return the string rater than having it in the trait.

src/engineio/transport_emitter.rs
nshaaban-cPacket: done https://github.com/1c3t3a/rust-socketio/pull/52/commits/c7ff0a0511af614b6243cf73ffc80b4ed93a9e52

nshaaban-cPacket:
For reviewing I'd recommend looking at https://github.com/1c3t3a/rust-socketio/pull/52/commits/1ff2a67c4bf183d692f692963ef49dfc8522ff61 , then going through each commit at a time (as all commits after that make small changes with readable diffs). Might want to turn off CI/CD for now, no reason to waste minutes on a work in progress.

nshaaban-cPacket:
Passes tests, but some performance issues with the SocketBuilder. Will fix after refactoring the remainder.

src/engineio/transports.rs
1c3t3a: Yeah that looks good!

nshaaban-cPacket:
Halting work pending your response on you want to proceed. @1c3t3a 

This PR is far too large. While there is more to do, it would be far easier to review (for you) if this was on a separate branch you could merge each non-breaking refactor PR individually. Do you want me to split this PR into chunks targeting a dev branch? (similar to what I have here https://github.com/nshaaban-cPacket/rust-socketio/pulls )

1c3t3a:
Ah I'm sorry! I'm planning to review it tomorrow (Saturday)! Actually just looking at the commit messages this looks like no part of the crate is left unchanged!

1c3t3a (commented):
Oke I looked through those changes and think that they are generally looking good. What kinda feels strange for me is the fact that nearly no architectural code is untouched, but as said, the changes look reasonable to me and also the refactoring of the socket.io module is fine overall. But a more in depth reviewing is necessary and I think we should therefore indeed split the refactoring into at least two PRs. One for the engine-io layer and one for socket.io. I could then review each part in depth and test it on my machine. Do you think it's possible for you to split the changes? I would open a `refactoring` branch which would serve as an integration location for all your changes step by step. Is that alright with you? Then I could also comment more critically about each change without needing to keep up with the sum of all other changes.

nshaaban-cPacket:
This will be split into smaller PRs. Closing...

PR #55, opened 2021-07-28
0.3.0 handshake packet

nshaaban-cPacket:
This will be used in socket later.

### DIFF
--- a/src/engineio/packet.rs
+++ b/src/engineio/packet.rs
@@ -57,7 +57,7 @@ pub struct HandshakePacket {
 impl TryInto<HandshakePacket> for Packet {
     type Error = Error;
     fn try_into(self) -> Result<HandshakePacket> {
-        Ok(serde_json::from_slice::<HandshakePacket>(self.data[..].as_ref())?)
+        Ok(serde_json::from_slice(self.data[..].as_ref())?)
     }
 }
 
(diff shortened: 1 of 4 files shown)


1c3t3a:
Alright! Could you please rebase so that only the relevant changes are shown?

src/engineio/packet.rs
```diff
@@ -41,6 +43,24 @@ const fn u8_to_packet_id(b: u8) -> Result<PacketId> {
...
+    #[serde(rename = "pingInterval")]
+    pub ping_interval: u64,
+    #[serde(rename = "pingTimeout")]
+    pub ping_timeout: u64,
+}
+
+impl TryInto<HandshakePacket> for Packet {
+    type Error = Error;
+    fn try_into(self) -> Result<HandshakePacket> {
+        Ok(serde_json::from_slice::<HandshakePacket>(self.data[..].as_ref())?)
```
1c3t3a, line 60: I think we don't need a turbofish `::<>`here anymore, type inference will catch this.

1c3t3a (changes requested):
Looks good, only one thing the rest is fine!

src/engineio/packet.rs
nshaaban-cPacket: done.

1c3t3a (approved):

PR #56, opened 2021-07-28
0.3.0 socket

nshaaban-cPacket:
Moved contents of transport into socket. While there are more changes to be made in socket, it needs to be moved first before diffs are readable.


src/engineio/socket.rs
```diff
@@ -1,22 +1,86 @@
...
+};
+use websocket::{
+    client::sync::Client as WsClient,
+    header::Headers,
+    sync::stream::{TcpStream, TlsStream},
+    ClientBuilder as WsClientBuilder,
+};
+use websocket::{
+    dataframe::Opcode, receiver::Reader, sync::Writer, ws::dataframe::DataFrame, Message,
+};
```
1c3t3a, line 29: Can't we union those imports? Or those the linter then starts to complain?

1c3t3a (approved):
Looks good

src/engineio/socket.rs
nshaaban-cPacket: done.

nshaaban-cPacket:
> Looks good! I think this one should be merged before the others, am I right?

Yup! Awaiting your merge, I don't have permission.

PR #57, opened 2021-07-28
0.3.0 create transports

nshaaban-cPacket:
Will be used by socket.rs

transport_new needs to be renamed after https://github.com/1c3t3a/rust-socketio/pull/56 is merged


src/engineio/transport_new.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// Used to update the base path, like when adding the sid.
+    fn set_base_url(&self, base_url: String) -> Result<()>;
+
+    /// Full query address
+    fn address(&self) -> Result<String> {
+        let reader = format!("{:#?}", SystemTime::now());
+        let hash = adler32(reader.as_bytes()).unwrap();
+        Ok(self.base_url()? + "&t=" + &hash.to_string())
+    }
```
1c3t3a, line 30: Maybe `get_address` is a better name, what do you think?

src/engineio/transport_new.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn emit(&self, data: Bytes, is_binary_att: bool) -> Result<()>;
+
+    /// Performs the server long polling procedure as long as the client is
+    /// connected. This should run separately at all time to ensure proper
+    /// response handling from the server.
+    fn poll(&self) -> Result<Bytes>;
+
+    /// Returns start of the url. ex. http://localhost:2998/engine.io/?EIO=4&transport=polling
+    /// Must have EIO and transport already set.
+    fn base_url(&self) -> Result<String>;
```
1c3t3a, line 20: I think `get_base_url` sounds better (as we also have the corresponding setter).

src/engineio/transport_new.rs
```diff
@@ -0,0 +1,52 @@
...
+        let bytes = Bytes::copy_from_slice(buf);
+        self.emit(bytes.clone(), true).unwrap();
+        Ok(bytes.len())
+    }
+
+    fn flush(&mut self) -> std::result::Result<(), std::io::Error> {
+        // We are always flushed.
+        Ok(())
+    }
+}
```
1c3t3a, line 52: Nice that's a lean way to do it!

1c3t3a (commented):
Yep that looks reasonable so far. Maybe we could try to get the base address to be a `&'_ str`, to get rid of the clone, but that could be saved for later!

src/engineio/transport_new.rs
nshaaban-cPacket: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

src/engineio/transport_new.rs
nshaaban-cPacket: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

nshaaban-cPacket:
> Yep that looks reasonable so far. Maybe we could try to get the base address to be a `&'_ str`, to get rid of the clone, but that could be saved for later!

I've been adding //TODOs whenever there is something we need to get back to later. In base_url's case it will be a Url (new struct) eventually (added a todo for it).

src/engineio/transport_new.rs
1c3t3a: Alright good catch!

nshaaban-cPacket:
Tests will be added later once EngineIoSocket is generic enough to lock it to one transport

1c3t3a (approved):
Looks good!

PR #58, opened 2021-07-28
0.3.0 socketio socket move

nshaaban-cPacket:
Moved socket out of lib.rs to it's own file.


1c3t3a (approved):
Yep looks good!

1c3t3a:
Ahh wait I just spotted an odd thing while looking at the coverage. Currently those lines:
https://github.com/1c3t3a/rust-socketio/blob/985305198a0148362148db329873fd9ca438ef41/src/lib.rs#L147-L159 

Are not covered meaning the ack did not work. I don't think that's too bad as the next changes would probably bring coverage back for those lines, but I found it a bit odd. What do you think @nshaaban-cPacket?

nshaaban-cPacket:
> Ahh wait I just spotted an odd thing while looking at the coverage. Currently those lines:
> 
> https://github.com/1c3t3a/rust-socketio/blob/985305198a0148362148db329873fd9ca438ef41/src/lib.rs#L147-L159
> 
> Are not covered meaning the ack did not work. I don't think that's too bad as the next changes would probably bring coverage back for those lines, but I found it a bit odd. What do you think @nshaaban-cPacket?

Likely the 2 second wait was up before the server could send the awk. I upped the timeout to 10, that should help. After we get the non-callback cfg we should transition the tests to that (as the tests should make sure the callback was actually used).

1c3t3a:
> > Ahh wait I just spotted an odd thing while looking at the coverage. Currently those lines:
> > https://github.com/1c3t3a/rust-socketio/blob/985305198a0148362148db329873fd9ca438ef41/src/lib.rs#L147-L159
> > 
> > Are not covered meaning the ack did not work. I don't think that's too bad as the next changes would probably bring coverage back for those lines, but I found it a bit odd. What do you think @nshaaban-cPacket?
> 
> Likely the 2 second wait was up before the server could send the awk. I upped the timeout to 10, that should help. After we get the non-callback cfg we should transition the tests to that (as the tests should make sure the callback was actually used).

Yep that makes sense. The CI pipeline somehow didn't run, I'll try to start it manually to see if this works after the time increasement.

nshaaban-cPacket:
@1c3t3a Ready for merge

PR #60, opened 2021-07-29
Added keygen script

nshaaban-cPacket:
Not sure if this is wanted, replaced the static keys (used in secure engine server in the dev environment) with a script to generate them. Keeps secrets out of the repo, but new contributors will need to run keygen (if they don't use make).

Ignore the GitGuardian check, it sees a secret and freaks out (even though the secret was removed in this commit)


ci/keygen.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/sh
+cd $(dirname $0)
+
+if [ "$1" = "" ] || [ "$2" = "" ]
+then
+    echo "Usage: keygen.sh [DOMIAN] [IP]"
```
1c3t3a, line 6: [DOMAIN]

Makefile
```diff
@@ -4,9 +4,11 @@ build:
 	@cargo build --verbose
 
 test-fast:
+	@./ci/keygen.sh node-engine-io-secure 127.0.0.1
 	@cargo test --verbose --package rust_socketio --lib -- engineio::packet && cargo test --verbose --package rust_socketio --lib -- socketio::packet
 
 test-all:
+	@./ci/keygen.sh node-engine-io-secure 127.0.0.1
```
1c3t3a, line 11: Maybe we could add a proper make target for the keys. Then the whole script is never called when the keys are present. The make target could then be invoked by the Github action instead of invoking it directly.

1c3t3a (changes requested):
I generally like the idea of it. But definitely something we should mention in the docs for new contributors.

ci/keygen.sh
nshaaban-cPacket: https://github.com/1c3t3a/rust-socketio/pull/65

Makefile
nshaaban-cPacket: done.

1c3t3a (approved):

PR #61, opened 2021-07-29
Into rather than encode

nshaaban-cPacket:
Use From and Into rather than encode/decode

### DIFF
--- a/src/engineio/packet.rs
+++ b/src/engineio/packet.rs
@@ -4,6 +4,7 @@ use bytes::{BufMut, Bytes, BytesMut};
 use serde::{Deserialize, Serialize};
 use std::char;
 use std::convert::TryInto;
+use std::convert::TryFrom;
 
 use crate::error::{Error, Result};
 /// Enumeration of the `engine.io` `Packet` types.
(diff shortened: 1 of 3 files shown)


src/engineio/socket.rs
```diff
@@ -609,7 +611,7 @@ impl EngineSocket {
                     PacketId::Message => {
                         let on_data = self.on_data.read()?;
                         if let Some(function) = on_data.as_ref() {
-                            spawn_scoped!(function(packet.data));
+                            spawn_scoped!(function(packet.data.clone()));
```
1c3t3a, line 614: Mhh how comes we're needing a clone here? Couldn't we just use `into_iter` here? For that case it might also make sense to implement `Iter` for `Payload`.

src/engineio/packet.rs
```diff
@@ -101,66 +117,77 @@ impl Packet {
...
-    // Observed some strange behavior while doing this with socket.io
-    // packets, works with engine.io packets.
-    /// Encodes a `Packet` with the payload as `base64`.
-    #[allow(dead_code)]
-    #[inline]
-    fn encode_base64(self) -> Bytes {
-        assert_eq!(self.packet_id, PacketId::Message);
+pub struct Payload {
+    packets: Vec<Packet>,
+}
```
1c3t3a, line 138: Does this payload struct clashes with the current `socketio::payload::Payload` struct?

nshaaban-cPacket: Shouldn't, its in a different module.

src/engineio/socket.rs
nshaaban-cPacket: Done and done

1c3t3a (approved):
Looks good now!

1c3t3a:

1c3t3a:
I would merge this after the conflicts are fixed ;)

PR #63, opened 2021-07-30
Send messages back to the client

nshaaban-cPacket:
To enable better testing


1c3t3a (approved):
Yep that definitely makes sense!

PR #64, opened 2021-07-30
Read env from mod.rs

nshaaban-cPacket:
Minimize repeated code in all tests


Cargo.toml
```diff
@@ -24,3 +24,4 @@ byte = "0.2.4"
 websocket = "0.26.2"
 thiserror = "1.0"
 native-tls = "0.2.7"
+url = "2.2.2"
```
1c3t3a, line 27: Could we skip this dependency by using either `Reqwest`s or `websocket`s URL type? If not that's completely fine, I just wondered.

nshaaban-cPacket: Reqwest and websocket's URL types are a re-export of url's. This also allows us to have socket.rs not reference either Reqwest or Websocket directly (the individual transports should do all the communication with those libraries. 
https://github.com/seanmonstar/reqwest/blob/e6a1a09f0904e06de4ff1317278798c4ed28af66/src/lib.rs#L223
https://github.com/websockets-rs/rust-websocket/blob/b78d08b9bdb401a5ddfb5bea97240af7c49e9795/src/lib.rs#L56

1c3t3a: Ah I see, makes sense!

1c3t3a (approved):

1c3t3a:

1c3t3a:
Mho it seems like some of the websocket tests time out after a while. What do you think, is it something that gets fixed within the next commits or do you plan on fixing it now?

nshaaban-cPacket:
> Mho it seems like some of the websocket tests time out after a while. What do you think, is it something that gets fixed within the next commits or do you plan on fixing it now?

Planning on refactoring tests next. Going to use some arcs and atomic bools to make sure the appropriate callbacks actually get called rather than just registered successfully. 

https://github.com/1c3t3a/rust-socketio/pull/61Is also stalled on passing tests.

nshaaban-cPacket:
This PR is priority \#1, without it the other PRs fail tests.

PR #66, opened 2021-07-30
Refactored engine.io socket

nshaaban-cPacket:


1c3t3a:

src/engineio/socket.rs
```diff
@@ -79,80 +37,153 @@ pub struct EngineSocket {
...
+    }
+
+    pub fn build(&self) -> Result<Socket<PollingTransport>> {
+        let (handshake, url) = self._build()?;
+
+        // Make a polling transport with new sid
+        let transport = PollingTransport::new(url, self.tls_config.clone(), self.headers.clone());
+
+        // If we can't upgrade or upgrade fails use polling
+        Ok(Socket::new(transport, handshake))
```
1c3t3a, line 99: If build was a consuming method (taking parameter `self`), we could get rid of all those `clone` statements on the member variables.

1c3t3a (approved):
Looks good!

src/engineio/socket.rs
nshaaban-cPacket: I was only able to get rid of one of the clone statements (as the shared _build function cannot consume self) by making them consume self. Changed the build methods to consume self so we are free to make more improvements in the future.

nshaaban-cPacket: refactored a bit more was able to remove another clone

nshaaban-cPacket:
Merge this one first, higher priority than the others.

Issue #67, opened 2021-08-04
Requests only going through sometimes?


Jefff000:
**For some reason, requests only go through sometimes and with little consistency. Im not too sure why? Maybe because its upgrading but Im a bit stuck...**

**Client-side Code**
```
use rust_socketio::{Payload, Socket, SocketBuilder};
use serde_json::json;

fn main() {
    let callback = |payload: Payload, mut socket: Socket| {
        match payload {
            Payload::String(str) => println!("Received: {}", str),
            Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
        }
        socket
            .emit("message", json!("aquired"))
            .expect("Server unreachable")
    };

    let mut socket = SocketBuilder::new("https://censored.herokuapp.com")
        .on("message", callback)
        .on("error", |err, _| eprintln!("Error: {:#?}", err))
        .connect()
        .expect("Connection failed");

    socket.emit("message", json!("cool message one")).expect("Server unreachable");
    socket.emit("message", json!("cool message two")).expect("Server unreachable");
    
    socket.emit("message", json!("cool message three")).expect("Server unreachable");
    socket.emit("message", json!("cool message four")).expect("Server unreachable");
}

```
**Serverside Code**
```
const cors = require('cors');
const app = require('express')();
const http = require('http').Server(app);
const io = require('socket.io')(http, {
	allowEIO3: true,
	cors: {
		origin: "*"
	}
});

const port = process.env.PORT || 3000;

io.on('connection', (socket) => {
	console.log("user connect");
	
	socket.on('message', function(data) {
		io.emit('message', data);
		console.log(data);
	});

	socket.on('clear', function clear() {
		io.emit('clear');
	});
});

http.listen(port, () => {
	console.log(`Socket.IO server running at http://localhost:${port}/`);
});

```

https://user-images.githubusercontent.com/83432863/128225249-f433997c-d36f-4ac3-aabb-8c5c721b09f9.mp4

nshaaban-cPacket:
Is this reproducible using http (rather than https)? @Jefff000

1c3t3a:
Could you run the server with DEBUG=1? Then we see which Transport Layer is used!

Jefff000:
[shoutbox-nullus-logs-1628112538106.txt](https://github.com/1c3t3a/rust-socketio/files/6934495/shoutbox-nullus-logs-1628112538106.txt)

Here is a log

Jefff000:
[...]

PR #68, opened 2021-08-04
Add debug to node servers

nshaaban-cPacket:
Inspired by https://github.com/1c3t3a/rust-socketio/issues/67#issuecomment-892971548 .

### DIFF
--- a/.devcontainer/docker-compose.yaml
+++ b/.devcontainer/docker-compose.yaml
@@ -6,18 +6,24 @@ services:
         command: [ "node", "/test/engine-io-secure.js" ]
         ports:
             - "4202:4202"
+        environment:
+            - "DEBUG=*"
     node-engine-io:
         build:
             context: ../ci
         command: [ "node", "/test/engine-io.js" ]
         ports:
             - "4201:4201"
+        environment:
+            - "DEBUG=*"
     node-socket-io:
         build:
             context: ../ci
         command: [ "node", "/test/socket-io.js" ]
         ports:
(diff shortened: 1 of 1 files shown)


1c3t3a:
Ah good catch! But I generally always start the CI servers with the command `DEBUG=*`, so that environment variable doesn't need to be set before (see: https://github.com/1c3t3a/rust-socketio/blob/5c66a6e755ed671eb0daabba596322653b7df5ff/ci/start_test_server.sh#L1-L24).

nshaaban-cPacket:
> Ah good catch! But I generally always start the CI servers with the command `DEBUG=*`, so that environment variable doesn't need to be set before (see:
> 
> https://github.com/1c3t3a/rust-socketio/blob/5c66a6e755ed671eb0daabba596322653b7df5ff/ci/start_test_server.sh#L1-L24
> ).

This PR sets the environment in the devcontainer (which doesn't use start_test_server.sh)

1c3t3a (approved):

1c3t3a:
> > Ah good catch! But I generally always start the CI servers with the command `DEBUG=*`, so that environment variable doesn't need to be set before (see:
> > https://github.com/1c3t3a/rust-socketio/blob/5c66a6e755ed671eb0daabba596322653b7df5ff/ci/start_test_server.sh#L1-L24
> > 
> > ).
> 
> This PR sets the environment in the devcontainer (which doesn't use start_test_server.sh)

Ahh I see!

PR #71, opened 2021-08-05
Main benchmarking

nshaaban-cPacket:
Adds bench-marking targets so performance before and after refactor can be compared. This does have to make the engineio module public in order to test. @1c3t3a Thoughts?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,10 @@ byte = "0.2.4"
 websocket = "0.26.2"
 thiserror = "1.0"
 native-tls = "0.2.7"
+
+[dev-dependencies]
+criterion = "0.3"
+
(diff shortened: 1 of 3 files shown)


nshaaban-cPacket:
> Mhh... having benchmarks before and after refactoring defenetly makes sense! But this pressures us to stabilize engine.io's API before releasing a new version, as then engine.io would be public. I'd say the best thing to do is, that we merge this PR, merge all of refactoring and then move the engine.io part into an own folder with it's own crate respectively. That crate then doesn't get published on crates.io until we decide that the API is stable enough.

That makes sense for 0.3.0 for now we can keep it on this branch so the benchmarks are still accessible and master has a private engineio

PR #79, opened 2021-08-12
feat: TransportTypes

nshaaban-cPacket:
Trait objects are a royal pain to deal with.

### DIFF
--- a/src/engineio/socket.rs
+++ b/src/engineio/socket.rs
@@ -1,9 +1,6 @@
-use crate::engineio::transport::Transport;
+use crate::engineio::transport::{Transport, TransportType};
 
-use super::transports::{
-    polling::PollingTransport, websocket::WebsocketTransport,
-    websocket_secure::WebsocketSecureTransport,
-};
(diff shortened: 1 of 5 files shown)


1c3t3a:
Mh I could understand that. Am I right that you plan on removing them completely?

nshaaban-cPacket:
> Mh I could understand that. Am I right that you plan on removing them completely?

When reasonable. There are times when trait objects can be really nice, however in this instance since the list of implementations is fixed an enum works much cleaner. It is difficult (if not impossible) to get trait objects Clonable, Sized, and usable.

1c3t3a:
> > Mh I could understand that. Am I right that you plan on removing them completely?
> 
> When reasonable. There are times when trait objects can be really nice, however in this instance since the list of implementations is fixed an enum works much cleaner. It is difficult (if not impossible) to get trait objects Clonable, Sized, and usable.

Alright, understand that, then lets keep it for now!

1c3t3a:
Could you resolve the conflicts? Then we're ready to merge!

PR #80, opened 2021-08-12
feat: improve test coverage

nshaaban-cPacket:


src/engineio/transports/polling.rs
```diff
@@ -87,34 +87,55 @@ impl Transport for PollingTransport {
...
-        assert_ne!(transport.base_url()?, url);
-        /*
-        //TODO: Make sure it doesn't double up
-        transport.set_base_url("127.0.0.1/?transport=polling".to_owned())?;
-        assert_eq!(transport.base_url()?, "127.0.0.1/?transport=polling");
-        assert_ne!(transport.base_url()?, url);
-        */
+        println!("{:?}", transport);
+        let test: Box<dyn Transport> = Box::new(transport);
+        println!("{:?}", test);
```
1c3t3a, line 138: Why you're looking for a debug test here? I don't really understand why there are the printlns

src/engineio/transports/polling.rs
```diff
@@ -87,34 +87,55 @@ impl Transport for PollingTransport {
     }
 
     fn set_base_url(&self, base_url: Url) -> Result<()> {
-        *self.base_url.write()? = base_url;
+        let mut url = base_url;
+        if url
+            .query_pairs()
+            .filter(|(k, v)| k == "transport" && v == "polling")
+            .count()
+            == 0
```
1c3t3a, line 95: You could use `any` here which should be faster as it returns whenever it found an occurrence of the thing you're looking at, other that `filter`+ `count` which looks at every element.

src/engineio/transports/websocket.rs
```diff
@@ -109,31 +109,51 @@ impl Transport for WebsocketTransport {
...
+        let mut url = url;
+        if url
+            .query_pairs()
+            .filter(|(k, v)| k == "transport" && v == "websocket")
+            .count()
+            == 0
+        {
+            url.query_pairs_mut().append_pair("transport", "websocket");
+        }
+        url.set_scheme("ws").unwrap();
```
1c3t3a, line 121: The same as above ;)

src/engineio/transports/websocket_secure.rs
```diff
@@ -114,27 +114,56 @@ impl Transport for WebsocketSecureTransport {
...
+        let mut url = url;
+        if url
+            .query_pairs()
+            .filter(|(k, v)| k == "transport" && v == "websocket")
+            .count()
+            == 0
+        {
+            url.query_pairs_mut().append_pair("transport", "websocket");
+        }
+        url.set_scheme("wss").unwrap();
```
1c3t3a, line 126: Same as above ;)

1c3t3a (changes requested):

src/engineio/transports/polling.rs
nshaaban-cPacket: Good optimization!

src/engineio/transports/polling.rs
nshaaban-cPacket: Woops, forgot to implement the actual test. https://github.com/1c3t3a/rust-socketio/pull/80/commits/097ee02bd141e6cc87b14a79218efe7761cdadfc

1c3t3a (approved):

PR #81, opened 2021-08-13
feat!: namespace return self

nshaaban-cPacket:
> That's nice! I always stumbled across the `Result` returned by the `set_namespace` method, that one's definitely better!
- https://github.com/1c3t3a/rust-socketio/pull/77#pullrequestreview-729576506

Good catch the type doesn't need to be result anymore.


1c3t3a (approved):
Great!

PR #85, opened 2021-08-13
0.2.X deprecation

nshaaban-cPacket:
This will be the final update to 0.2.X before 0.3.0 goes live. Uses deprecation notes to walk the user through how to upgrade from 0.2.X to 0.3.0. AKA if the users resolve all deprecation warnings described in the notes they can upgrade to 0.3.0 without errors. (even without looking at the changelog)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,7 +177,7 @@ impl SocketBuilder {
 
     /// Sets the target namespace of the client. The namespace must start
     /// with a leading `/`. Valid examples are e.g. `/admin`, `/foo`.
-    #[deprecated(note = "use namespace instead")]
+     #[deprecated(note = "use namespace(nsp) instead")]
     pub fn set_namespace<T: Into<String>>(mut self, namespace: T) -> Result<Self> {
         let nsp = namespace.into();
         if !nsp.starts_with('/') {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,7 +177,7 @@ impl SocketBuilder {
 
     /// Sets the target namespace of the client. The namespace must start
     /// with a leading `/`. Valid examples are e.g. `/admin`, `/foo`.
-     #[deprecated(note = "use namespace(nsp) instead")]
+    #[deprecated(note = "use namespace(nsp) instead")]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -177,6 +177,7 @@ impl SocketBuilder {
 
     /// Sets the target namespace of the client. The namespace must start
     /// with a leading `/`. Valid examples are e.g. `/admin`, `/foo`.
+    #[deprecated(note = "use namespace instead")]
```
1c3t3a, line 180: We might write 'use `namespace(nsp)` instead' in order to indicate that this refers to a function.

src/lib.rs
```diff
@@ -239,11 +247,37 @@ impl SocketBuilder {
     ///     .connect();
     ///
     /// ```
+    #[deprecated(note = "use tls_config instead")]
```
1c3t3a, line 250: Same like the nsp thing above.

src/lib.rs
```diff
@@ -261,6 +295,7 @@ impl SocketBuilder {
     ///     .connect();
     ///
     /// ```
+    #[deprecated(note = "use opening_header instead")]
```
1c3t3a, line 298: Same as above

1c3t3a (commented):
This generally looks good with just a few points. Do you think it makes sense to add a `since = "0.3.0"` to the deprecation macros?

src/lib.rs
nshaaban-cPacket:
```suggestion
     #[deprecated(note = "use namespace(nsp) instead")]
```

src/lib.rs
nshaaban-cPacket:
```suggestion
     #[deprecated(note = "use opening_header(header, value) instead")]
```

src/lib.rs
nshaaban-cPacket:
```suggestion
     #[deprecated(note = "use tls_config(tls_config) instead")]
```

1c3t3a (approved):
Alright, looks good!

PR #87, opened 2021-08-13
feat: merged socket and transport

nshaaban-cPacket:


1c3t3a:
Mhh why would you do that? I think it's rather unhealthy to have a file with over 900 lines, isn't it?

nshaaban-cPacket:
> Mhh why would you do that? I think it's rather unhealthy to have a file with over 900 lines, isn't it?

Reasons to remove:
- Duplication of documentation (Socket.emits doc comment is virtually identical to that of TransportClient
- The "TransportClient" is more of a engineio socket wrapper+socketio features rather than a transport client or wrapper
- All of the functions in Socket are purely boilerplate (all TransportClient's public methods are wrapped)

Reasons to keep in separate files
- Allow easier modification of Socket's method headers without semver breaking changes.

Alternative solution to above issues (Will do this in a separate PR later):
- Rename TransportClient to something closer related to what it is
- Have TransportClient hold the client only behavior (giving it a justification for being in a separate file)

PR #93, opened 2021-08-23
feat!: split into two crates

nshaaban-cPacket:
Breaking changes:
- Unneeded type annotations. For example on https://github.com/1c3t3a/rust-socketio/blob/896a23d6cf4a972a33c365cf904a1c8abf4fbd4f/src/socketio/client/socket.rs#L155 the .parse().unwrap() is no longer necessary.
- Removed unused Error types


engineio/src/header.rs
```diff
@@ -0,0 +1,170 @@
...
+        Self::new()
+    }
+}
+
+impl Iterator for IntoIter {
+    type Item = (HeaderName, HeaderValue);
+    fn next(&mut self) -> std::option::Option<<Self as std::iter::Iterator>::Item> {
+        self.inner.next()
+    }
+}
```
1c3t3a, line 170: Nice! Better to have it this way!

1c3t3a (approved):
That looks exactly like discussed! Thank you ;)

PR #97, opened 2021-08-26
refactor: move callback init to builder

nshaaban-cPacket:
Refactor to make IllegalActionAfterOpen impossible in engineio. This is not a breaking change thanks to @1c3t3a having the wisdom not to make public yet. Thank you.

### DIFF
--- a/engineio/src/callback.rs
+++ b/engineio/src/callback.rs
@@ -1,0 +1,89 @@
+use crate::Packet;
+use bytes::Bytes;
+use std::fmt::Debug;
+use std::ops::Deref;
+use std::sync::Arc;
+
+pub(crate) type Callback<I> = dyn Fn(I) + 'static + Sync + Send;
+
+#[derive(Clone)]
+/// Internal type, only implements debug on fixed set of generics
+pub(crate) struct OptionalCallback<I> {
+    inner: Arc<Option<Box<Callback<I>>>>,
+}
+
+impl<I> OptionalCallback<I> {
+    pub(crate) fn new<T>(callback: T) -> Self
+    where
+        T: Fn(I) + 'static + Sync + Send,
+    {
+        OptionalCallback {
+            inner: Arc::new(Some(Box::new(callback))),
+        }
+    }
+
+    pub(crate) fn default() -> Self {
+        OptionalCallback {
+            inner: Arc::new(None),
+        }
+    }
+}
+
+impl Debug for OptionalCallback<String> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        f.write_fmt(format_args!(
+            "Callback({:?})",
+            if self.inner.is_some() {
+                "Fn(String)"
+            } else {
+                "None"
+            }
+        ))
+    }
+}
+
+impl Debug for OptionalCallback<()> {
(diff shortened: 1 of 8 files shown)


engineio/src/callback.rs
```diff
@@ -0,0 +1,89 @@
+use crate::Packet;
+use bytes::Bytes;
+use std::fmt::Debug;
+use std::ops::Deref;
+use std::sync::Arc;
+
+pub(crate) type Callback<I> = dyn Fn(I) + 'static + Sync + Send;
+
+#[derive(Clone)]
+/// Internal type, only implements debug on fixed set of generics
```
1c3t3a, line 10: That's nice!

engineio/src/client/socket.rs
```diff
@@ -32,19 +38,72 @@ impl SocketBuilder {
             headers: None,
             tls_config: None,
             handshake: None,
+            on_close: OptionalCallback::default(),
```
1c3t3a, line 41: Didn't you plan on putting those in a map? Or is that something that happens in a separate pr?

engineio/src/client/socket.rs
```diff
@@ -32,19 +38,72 @@ impl SocketBuilder {
...
+    /// Registers the `on_packet` callback.
+    pub fn on_packet<T>(mut self, callback: T) -> Self
+    where
+        T: Fn(Packet) + 'static + Sync + Send,
+    {
+        self.on_packet = OptionalCallback::new(callback);
+        self
+    }
+
+    /// Preforms the handshake
```
1c3t3a, line 106: Performs

engineio/src/client/socket.rs
1c3t3a: Ah wait that's the builder, nevermind!

1c3t3a (commented):
LGTM

engineio/src/client/socket.rs
nshaaban-cPacket: I attempted to put them both in a map and struct, opted to go for the way it was.
Map issues:
- The generic type varies between the callbacks, could use a dyn object to try to capture this, that has it's own issues
Struct issues:
- No real value add
- In order to retrieve the values from the struct the function name to get the callback out (on_close or close) taking no parameters looks more (from a socketio background) like a function that would call the events or handlers, not simple getters
- Difficult to get type checking on enclosures to resolve correctly due to lack of trait aliases in rust

engineio/src/client/socket.rs
nshaaban-cPacket: 961d1fb3a6f7a45a474966c1cd64ae83118b9faa

engineio/src/client/socket.rs
1c3t3a: Ah oke I See! Didn't think about the generic type parameter. The type parameter could be refactored into an enum, but I'm not sure if that brings some enhancement or just makes things more difficult. I think the current way is pretty straight forward.

1c3t3a (approved):
Just some merge conflicts and we're ready to go!

nshaaban-cPacket:
Sending this PR into Draft as it(?)'s intermittently failing with "invalid namespace" https://github.com/1c3t3a/rust-socketio/runs/3465528073?check_suite_focus=true https://github.com/1c3t3a/rust-socketio/runs/3465739096?check_suite_focus=true

Issue #94, opened 2021-08-25
Limit transport to websockets


namxam:
Is it possible to specify the transport before initializing a connection? We have a service which only supports websocket connections and no long polling.

1c3t3a:
That's currently not supported, but a feature that's rather easy to add. There is currently a lot of refactoring going on by @nshaaban-cPacket, but I'm sure we could add this to the list of features. Currently the first request is a http request to the server, receiving a handshaking packet in response. If this packet mentions an upgrade possibility, the client upgrades immediately.

1c3t3a:
How urgent is your need for this? We could maybe add this before working on a server implementation after the first refactoring PRs hit main.

nshaaban-cPacket:
@namxam released in 0.3.0-alpha-1 https://docs.rs/rust_socketio/0.3.0-alpha.1/rust_socketio/struct.SocketBuilder.html#method.transport_type

1c3t3a:
Closing this as it is now resolved.

PR #98, opened 2021-08-26
feat: pure websocket build function

ctrlaltf24:
Engineio half of #94 . Adds build functions to create websocket transports without relying on polling.


1c3t3a (approved):
LGTM!

PR #99, opened 2021-08-26
0.3.0 socketio specify transport types

nshaaban-cPacket:
Depends on:
- https://github.com/1c3t3a/rust-socketio/pull/98

Socketio half of #94

### DIFF
--- a/socketio/src/client/socket.rs
+++ b/socketio/src/client/socket.rs
@@ -1,7 +1,9 @@
 pub use super::super::{event::Event, payload::Payload};
 use crate::packet::Packet as SocketPacket;
 use native_tls::TlsConnector;
+use rust_engineio::client::{Socket as EngineIoSocket, SocketBuilder as EngineIoSocketBuilder};
 use rust_engineio::header::{HeaderMap, HeaderValue};
+use url::Url;
 
 use crate::error::Result;
 use std::{time::Duration, vec};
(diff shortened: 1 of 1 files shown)


socketio/src/client/socket.rs
```diff
@@ -20,6 +22,16 @@ pub struct Socket {
 
 type SocketCallback = dyn FnMut(Payload, Socket) + 'static + Sync + Send;
 
+/// Flavor of engineio transport
+//TODO: Consider longevity
```
1c3t3a, line 26: What do you mean by that?

socketio/src/client/socket.rs
```diff
@@ -192,13 +214,29 @@ impl SocketBuilder {
...
+        }
+
+        let mut builder = EngineIoSocketBuilder::new(url);
+
+        if let Some(tls_config) = self.tls_config {
+            builder = builder.tls_config(tls_config);
+        }
+        if let Some(headers) = self.opening_headers {
+            builder = builder.headers(headers);
+        }
```
1c3t3a, line 230: Would it make sense to have `builder.headers()` or `builder.tls_config()` taking an `Option<T>`? That way we wouldn't have the if let here and the builder holds them as `Option` anyways.

1c3t3a: But it feels strange to call a setter with an optional value...

nshaaban-cPacket: > But it feels strange to call a setter with an optional value...

Ultimately our primary goal is to provide an easy to use interface to the consumers of this API, if that means we need to have a tiny bit more boilerplate to save the users from writing Some() around every call to the setters headers and tls_config so be it. In the future this EngineIO builder will also be public.

socketio/src/client/socket.rs
nshaaban-cPacket: Is storing the underlying transport type in an enum (with those options) what we want to use in the future? Moving or modifying (even adding) the enum in the future is a breaking change.

socketio/src/client/socket.rs
1c3t3a: Yep that's true.

Issue #90, opened 2021-08-16
Multiple Attachment Support


nshaaban-cPacket:
Seems like there is only a field for [one attachment](https://github.com/1c3t3a/rust-socketio/blob/d421946915b1aa6cf95b65624853ffd0717747b6/src/socketio/packet.rs#L24) (and [only one stored](https://github.com/1c3t3a/rust-socketio/blob/d421946915b1aa6cf95b65624853ffd0717747b6/src/socketio/transport.rs#L276)), despite [storing how many attachments](https://github.com/1c3t3a/rust-socketio/blob/d421946915b1aa6cf95b65624853ffd0717747b6/src/socketio/packet.rs#L26) a packet has. Was there a reason for this @1c3t3a or is it just not implemented yet?

TODO:
- [ ] Change binary data structure to a list
- [ ] Consume as many packets as required

1c3t3a:
It's actually not yet implemented. That's defenetly something that should be added. In the current setup this would lead to just having the last binary payload as binary_payload of a packet.

PR #101, opened 2021-08-27
0.3.0 socketio attachments

ctrlaltf24:
Breaking changes:
- socketio packet binary_data removed
- socketio packet attachments -> attachment_count
- socketio packet attachments added
- socketio packet attachment_count: u8
- socketio send takes ownership of packet

Added deprecations with notes in the 0.2.X deprecation branch

fixes: #90

### DIFF
--- a/socketio/src/socket.rs
+++ b/socketio/src/socket.rs
@@ -299,7 +299,7 @@ impl Socket {
     /// This method is later registered as the callback for the `on_data` event of the
     /// engineio client.
     #[inline]
-    fn handle_new_packet(&self, socket_packet: SocketPacket) -> Option<SocketPacket> {
+    fn handle_new_socketio_packet(&self, socket_packet: SocketPacket) -> Option<SocketPacket> {
         let output = socket_packet.clone();
 
         let default = String::from("/");
@@ -349,6 +349,50 @@ impl Socket {
             }
         }
         Some(output)
+    }
+
+    /// Handles new incoming engineio packets
+    pub fn handle_new_engineio_packet(
+        &self,
+        engine_iter: &mut rust_engineio::client::Iter,
+        packet: EnginePacket,
+    ) -> Option<Result<SocketPacket>> {
+        let socket_packet = SocketPacket::try_from(&packet.data);
+        if let Err(err) = socket_packet {
+            return Some(Err(err));
+        }
+        // SAFETY: checked above to see if it was Err
+        let mut socket_packet = socket_packet.unwrap();
+        // Only handle attachments if there are any
+        if socket_packet.attachment_count > 0 {
+            let mut attachments_left = socket_packet.attachment_count;
+            let mut attachments = Vec::new();
+            while attachments_left > 0 {
+                let next = engine_iter.next()?;
+                match next {
+                    Err(err) => return Some(Err(err.into())),
+                    Ok(packet) => match packet.packet_id {
+                        EnginePacketId::MessageBinary | EnginePacketId::Message => {
+                            attachments.push(packet.data);
+                            attachments_left = attachments_left - 1;
+                        }
+                        _ => {
+                            return Some(Err(Error::InvalidAttachmentPacketType(
+                                packet.packet_id.into(),
+                            )));
+                        }
+                    },
+                }
+            }
+            socket_packet.attachments = Some(attachments);
+        }
+
+        let packet = self.handle_new_socketio_packet(socket_packet);
+        if let Some(packet) = packet {
+            return Some(Ok(packet));
+        } else {
+            return None;
+        }
     }
 
     /// Handles the incoming acks and classifies what callbacks to call and how.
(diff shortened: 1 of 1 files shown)


engineio/src/socket.rs
```diff
@@ -117,7 +117,7 @@ impl Socket {
     }
 
     pub fn close(&self) -> Result<()> {
-        self.emit(Packet::new(PacketId::Close, Bytes::from_static(b"")), false)?;
+        self.emit(Packet::new(PacketId::Close, Bytes::from_static(b"")))?;
```
1c3t3a, line 120: I think `Bytes::default()` or `Bytes::new()` is better here. The [documentation](https://docs.rs/bytes/1.1.0/bytes/struct.Bytes.html#method.new) here also mentions that empty bytes also don't allocate.

1c3t3a: This involves all occurrences of a `Bytes::from_static(b"")` call.

socketio/src/packet.rs
```diff
@@ -117,6 +123,13 @@ impl From<&Packet> for Bytes {
     }
 }
 
+impl TryFrom<Bytes> for Packet {
+    type Error = Error;
+    fn try_from(value: Bytes) -> Result<Self> {
+        Packet::try_from(&value)
+    }
+}
```
1c3t3a, line 131: That's nice to have!

socketio/src/socket.rs
```diff
@@ -197,24 +193,16 @@ impl Socket {
...
-    /// should only be called if a `BinaryEvent` or `BinaryAck` was
-    /// send to the server mentioning this attachment in it's
-    /// `attachments` field.
-    fn send_binary_attachment(&self, attachment: Bytes) -> Result<()> {
-        if !self.is_engineio_connected()? || !self.connected.load(Ordering::Acquire) {
-            return Err(Error::IllegalActionBeforeOpen());
+        if let Some(ref attachments) = packet.attachments {
+            for attachment in attachments {
+                let engine_packet =
+                    EnginePacket::new(EnginePacketId::MessageBinary, attachment.clone());
```
1c3t3a, line 201: Why do we clone here? It should be possible to partially move ownership of the bytes here, what do you think?

socketio/src/socket.rs
```diff
@@ -322,90 +300,56 @@ impl Socket {
...
-            // this is a normal packet, so decode it
-            SocketPacket::try_from(&socket_bytes)
-        };
+    fn handle_new_packet(&self, socket_packet: SocketPacket) -> Option<SocketPacket> {
+        let output = socket_packet.clone();
 
-        if let Ok(socket_packet) = decoded_packet {
-            let output = socket_packet.clone();
+        let default = String::from("/");
+        //TODO: should nsp logic be here?
```
1c3t3a, line 307: Good question, where else does the packet get handled? Generally one could expect that such things happen in a method named `handle_new_packet`.

socketio/src/socket.rs
```diff
@@ -585,19 +533,44 @@ impl<'a> Iterator for Iter<'a> {
...
+                                EnginePacketId::MessageBinary | EnginePacketId::Message => {
+                                    attachments.push(packet.data);
+                                    attachments_left = attachments_left - 1;
+                                }
+                                _ => {
+                                    return Some(Err(Error::InvalidAttachmentPacketType(
+                                        packet.packet_id.into(),
+                                    )));
+                                }
+                            }
```
1c3t3a, line 565: This could be collided to a match block instead of unwrapping after the if let.

1c3t3a (changes requested):
Generally this looks really good, thanks a lot! Just a few things I had questions about ;)

socketio/src/packet.rs
nshaaban-cPacket: The existing choice to use a reference (as it doesn't require ownership) was fascinating, don't see that very often but makes sense!

socketio/src/socket.rs
nshaaban-cPacket: Good catch, the send function should take ownership of the packet rather than just a reference. 30675a2cce96b3c6db4f1f1df0b75764c1d859e7

socketio/src/socket.rs
nshaaban-cPacket: Packet lifetime (planned for 0.3.0):
- SocketIO client socket connect method
  - SocketIO common socket connect method
  - Separate thread loop
    - SocketIO client socket iterator
      - Socketio common socket iterator
        - EngineIO iterator
          - Transport
          - EngineIO parsing
          - EngineIO handling (like responding to pings)
        - SocketIO common socket parsing
        - SocketIO common socket handling (no client side logic)
      - SocketIO client socket handling (client side logic)
Upcoming in future PRs

socketio/src/socket.rs
nshaaban-cPacket: Originally had it like that, but tab level was getting out of hand. a03ad2f6ed020cd96e56eb53bec7314923d8d049.
Refactored into a handle_new_entineio_packet function in 6c3b48c04d82f60c00ec039823902fa45b5aeeb2
Will be refactoring this section further in PRs to come, out of scope for this PR (added to checklist).

engineio/src/socket.rs
nshaaban-cPacket: 17a394c9339f537772b689814feb091a487ee34d

socketio/src/socket.rs
1c3t3a: Yep, lets handle this in a future PR ;)

socketio/src/socket.rs
1c3t3a: Alright!

PR #103, opened 2021-09-03
feat: generic engineio socket

nshaaban-cPacket:
EngineIO Socket trait to allow for either future server engineio socket or client engineio socket in common socket.io socket

### DIFF
--- a/engineio/src/model.rs
+++ b/engineio/src/model.rs
@@ -1,7 +1,7 @@
 use crate::error::Result;
 use crate::Packet;
 
-pub trait Socket {
+pub trait Socket: Send + Sync + Clone {
     fn close(&self) -> Result<()>;
     fn connect(&self) -> Result<()>;
     fn emit(&self, packet: Packet) -> Result<()>;
--- a/engineio/src/model.rs
+++ b/engineio/src/model.rs
@@ -1,0 +1,9 @@
+use crate::error::Result;
+use crate::Packet;
+
+pub trait Socket {
+    fn close(&self) -> Result<()>;
+    fn connect(&self) -> Result<()>;
+    fn emit(&self, packet: Packet) -> Result<()>;
+    fn is_connected(&self) -> Result<bool>;
+}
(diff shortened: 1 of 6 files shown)


engineio/src/model.rs
```diff
@@ -0,0 +1,9 @@
+use crate::error::Result;
+use crate::Packet;
+
+pub trait Socket {
+    fn close(&self) -> Result<()>;
+    fn connect(&self) -> Result<()>;
+    fn emit(&self, packet: Packet) -> Result<()>;
+    fn is_connected(&self) -> Result<bool>;
```
1c3t3a, line 8: Nice! Everything with `&_` allows to just pass an `&Socket` to the callback!

socketio/src/socket.rs
```diff
@@ -41,8 +36,10 @@ pub struct Ack {
 
 /// Handles communication in the `socket.io` protocol.
 #[derive(Clone)]
-pub struct Socket {
-    engine_socket: Arc<EngineIoSocket>,
+pub struct Socket<
+    EngineSocket: rust_engineio::model::Socket + Send + Sync + Clone + 'static + Debug,
```
1c3t3a, line 40: Would it make sense to have those trait bounds directly on the Socket trait? I don't think there will be an implementation that's not Send + Sync + ...

1c3t3a (commented):
That looks really good! Very clean with the trait and the generic Type on the socket.

socketio/src/socket.rs
nshaaban-cPacket: 86d501ac1fcf7116630e3e8895b69616349eb89d

engineio/src/model.rs
nshaaban-cPacket: EngineIO doesn't pass the socket to engineio callbacks today, only socketio exposes a socket in the callbacks.

nshaaban-cPacket:
found a better way

PR #105, opened 2021-09-04
! 0.3.0 socketio callback refactor

nshaaban-cPacket:
Sorry about the mega PR, things got out of hand. Did manage to make the code 200 lines shorter.

This PR cleans up 
- https://github.com/1c3t3a/rust-socketio/pull/101#discussion_r698482795
- https://github.com/1c3t3a/rust-socketio/pull/101#discussion_r698477692

Other notes:
- Iterator should be usable by users now (so long as they connect_manual)
- Implementing trait based callbacks should be relatively easy (see callback function)
- Will require some refactoring internally to get server implementation to share code (see hard coded engineio socket client in common socketio socket)

Breaking changes:
- Event type on callback no longer needs into

### DIFF
--- a/engineio/src/client/socket.rs
+++ b/engineio/src/client/socket.rs
@@ -301,7 +301,7 @@ impl Socket {
     }
 
     /// Polls for next payload
-    //TODO: private?
+    #[doc(hidden)]
     pub fn poll(&self) -> Result<Option<Packet>> {
         let packet = self.socket.poll()?;
         if let Some(packet) = packet {
--- a/engineio/src/client/socket.rs
+++ b/engineio/src/client/socket.rs
@@ -4,7 +4,7 @@ use crate::transport::Transport;
 
 use crate::error::{Error, Result};
 use crate::header::HeaderMap;
-use crate::packet::{HandshakePacket, Packet, PacketId, Payload};
+use crate::packet::{HandshakePacket, Packet, PacketId};
 use crate::transports::{PollingTransport, WebsocketSecureTransport, WebsocketTransport};
 use bytes::Bytes;
 use native_tls::TlsConnector;
@@ -280,8 +280,8 @@ impl SocketBuilder {
 }
 
 impl Socket {
-    pub fn close(&mut self) -> Result<()> {
-        self.socket.close()
+    pub fn close(&self) -> Result<()> {
+        self.socket.disconnect()
     }
 
     /// Opens the connection to a specified server. The first Pong packet is sent
@@ -290,24 +290,21 @@ impl Socket {
         self.socket.connect()
     }
 
+    /// Disconnects the connection.
+    pub fn disconnect(&self) -> Result<()> {
+        self.socket.disconnect()
+    }
+
     /// Sends a packet to the server.
     pub fn emit(&self, packet: Packet) -> Result<()> {
         self.socket.emit(packet)
     }
 
     /// Polls for next payload
-    pub(crate) fn poll(&self) -> Result<Option<Payload>> {
-        let payload = self.socket.poll()?;
-
-        if payload.is_none() {
-            return Ok(None);
-        }
-
-        let payload = payload.unwrap();
-
-        let iter = payload.iter();
-
-        for packet in iter {
+    //TODO: private?
+    pub fn poll(&self) -> Result<Option<Packet>> {
+        let packet = self.socket.poll()?;
+        if let Some(packet) = packet {
             // check for the appropriate action or callback
             self.socket.handle_packet(packet.clone())?;
             match packet.packet_id {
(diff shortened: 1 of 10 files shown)


engineio/src/client/socket.rs
```diff
@@ -290,24 +290,21 @@ impl Socket {
...
-        if payload.is_none() {
-            return Ok(None);
-        }
-
-        let payload = payload.unwrap();
-
-        let iter = payload.iter();
-
-        for packet in iter {
+    //TODO: private?
```
1c3t3a, line 304: What do you think about `#[doc(hidden)]`? As far as I understood the poll method is needed by the socketio part, or am I wrong?

1c3t3a: Generally speaking: I think the iterator should be the go-to way of receiving packets as its nice and clean.

engineio/src/socket.rs
```diff
@@ -103,17 +113,37 @@ impl Socket {
...
+        loop {
+            if self.connected.load(Ordering::Acquire) {
+                if self.remaining_packets.read()?.is_some() {
+                    // SAFETY: checked is some above
+                    let mut iter = self.remaining_packets.write()?;
+                    let iter = iter.as_mut().unwrap();
+                    if let Some(packet) = iter.next() {
+                        return Ok(Some(packet));
+                    } else {
+                        // Iterator has run out of packets, get a new payload
```
1c3t3a, line 126: Is that a TODO?

1c3t3a (commented):
This generally looks good, thanks! The type inference issue sounds ugly... I think that could be a real issue as this is currently an example from documentation (if I remember correctly), where we could expect that this actually breaks stuff that's currently running in compile time. But I see that other alternatives also don't make the usage easier and I like that the clone is replaced by a reference. A possible, but still ugly alternative could be an `Arc<Socket>` as an argument (potentially in the flavor of a wrapper type). What is your opinion on that? Do you see any solution to it? I wouldn't expect the rust compiler to change the behavior here, so that's a quite big decision.

nshaaban-cPacket:
> This generally looks good, thanks! The type inference issue sounds ugly... I think that could be a real issue as this is currently an example from documentation (if I remember correctly), where we could expect that this actually breaks stuff that's currently running in compile time. But I see that other alternatives also don't make the usage easier and I like that the clone is replaced by a reference. A possible, but still ugly alternative could be an `Arc<Socket>` as an argument (potentially in the flavor of a wrapper type). What is your opinion on that? Do you see any solution to it? I wouldn't expect the rust compiler to change the behavior here, so that's a quite big decision.

## Static Wrapper 
```rust
type InnerCallback = Box<dyn for<'a> FnMut(Payload, SocketRef) + 'static + Sync + Send>;
pub struct SocketRef {
    inner: &'static Socket
}
```
Failed due to needing 'static for all references to self

## Lifetime wrapper
```rust
type InnerCallback = Box<dyn for<'a> FnMut(Payload, SocketRef<'a>) + 'static + Sync + Send>;
pub struct SocketRef<'a> {
    inner: &'a Socket
}
```
Failed for the same reason the reference failed.

> Do you see any solution to it?

Presently no, AFIK this is the cleanest way, unfortunate that the compiler's lifetime infrense was insufficient. At this point it's not worth the marginal improvement having callbacks take references. Reverted back to cloning, perhaps we can be more idiomatic in iterators/callback traits

engineio/src/client/socket.rs
nshaaban-cPacket: > What do you think about `#[doc(hidden)]`? As far as I understood the poll method is needed by the socketio part, or am I wrong?

done

nshaaban-cPacket: > Generally speaking: I think the iterator should be the go-to way of receiving packets as its nice and clean.

Made socketio iter pub

engineio/src/socket.rs
nshaaban-cPacket: ad5513e5ae6ce426e6851da23c9336664e32d2c0

1c3t3a:
> Presently no, AFIK this is the cleanest way, unfortunate that the compiler's lifetime infrense was insufficient. At this point it's not worth the marginal improvement having callbacks take references. Reverted back to cloning, perhaps we can be more idiomatic in iterators/callback traits

Yep that sounds reasonable. A pitty that even the other alternatives don't work and that type inference is presenting such a struggle. I hope that we'll find an alternative to `Fn` in the future anyways, especially as I think it brings in some problems when running in the async context.

nshaaban-cPacket:
> Yep that sounds reasonable. A pitty that even the other alternatives don't work and that type inference is presenting such a struggle. I hope that we'll find an alternative to `Fn` in the future anyways, especially as I think it brings in some problems when running in the async context.

Applied the reversion to this branch, this PR still makes the callback type a little cleaner and makes dataflow more consistent.

1c3t3a (approved):
> Applied the reversion to this branch, this PR still makes the callback type a little cleaner and makes dataflow more consistent.

That's true!

PR #107, opened 2021-09-07
Optimize websocket import to not compile tokio

1c3t3a:
This is a small change that makes sure to only compile the necessary bits of the websocket crate.


nshaaban-cPacket:
@1c3t3a The cargo.toml moved into the engineio directory (due to crate split)

1c3t3a:
Yep, I saw that. Should be fine now.

nshaaban-cPacket (approved):

PR #119, opened 2021-09-14
0.3.0 invalid namespace

nshaaban-cPacket:
Problem was two-fold:
- no socket.of("/admin").on("connection", callback) in ci
- Integration tests didn't use namespaces other than /


1c3t3a (approved):
Thanks!

Issue #115, opened 2021-09-14
Switch websocket library


1c3t3a:
This library currently uses [the websocket crate](https://github.com/websockets-rs/rust-websocket/). However the library looks kind of abandoned and might not be the best fit for future plans like `async / await` support. Currently the library provides async support based on `tokio` 0.1. This could lead to problems and there are no real plans to get the support running with websocket (see [#242](https://github.com/websockets-rs/rust-websocket/issues/242)). It's also not to be expected that the library will provide support in case we have a problem, so switching the library might be a good idea. A battle proven implementation is provided by `tungstenite-rs` (async [here](https://github.com/snapview/tokio-tungstenite), blocking [here](https://github.com/snapview/tungstenite-rs)). What do you think @nshaaban-cPacket?

nshaaban-cPacket:
Sounds good, might be worth doing before the engineio crate goes public? I'm not interested in implementing the change in the near future (only if it causes headache with 0.5.0).

1c3t3a:
Alright, I could implement it as soon as refactoring is merged. Also a good point that this should be done before engine.io goes public, then I am not tied to a certain API.

nshaaban-cPacket:
> Alright, I could implement it as soon as refactoring is merged. Also a good point that this should be done before engine.io goes public, then I am not tied to a certain API.

Doesn't necessary have to be before, but might make sense. Ideally what library we are using internally for websockes/polling would not be exposed anyways.

nshaaban-cPacket:
websocket *is* exposed here https://github.com/1c3t3a/rust-socketio/blob/9b1183b0313b52b82bb9e09f7c75f223cb9958ee/engineio/src/error.rs#L44

Plus we'll need the https://docs.rs/tungstenite/0.15.0/tungstenite/fn.client_tls_with_config.html to continue supporting passing in a TlsConnector

nshaaban-cPacket:
May want to escalate this @1c3t3a websocket (latest) depends on hyper 0.10.6 which is vulnerable to RUSTSEC-2021-0079 and RUSTSEC-2021-0078 (found using cargo-audit)

vi:
`websocket` crate is basically unmaintained now. Switching to `tungstenite` is recommended.

---

[...]

PR #123, opened 2021-10-01
feat: switch to tungstenite-rs

nshaaban-cPacket:
Was going to get in the way, plus has more breaking changes (see changelog changes).

Fixes: #115


nshaaban-cPacket:
tungstenite-rs is a blocking library (that doesn't support .split). Observed significant performance degradation (25s minimum, peak of 75s) in the following case (client::socket::test::socket_io_integration):
- Callback mode (where there is a separate thread constantly calling poll)
- Attempt to send packets on the clone of the socket returned
- Observe that websocket connection is busy reading
- When a ping packet is received it is **possible** that the lock switches to the write
- Socket finally sends a message

(keep in mind this same performance degredation is likely on the secure websocket connection as it also doesn't support .split)

Suggested fix: Use some async internally. Currently we are trying to expose a callback interface (typically async), yet only using blocking libraries. Delay until 0.4.0, as it introduces a general performance degradation.

1c3t3a:
Superseded by #146.

PR #125, opened 2021-10-04
feat!: deprication cleanup

nshaaban-cPacket:
Breaking changes:
- Removed headermap from pub use, internal type only
- hide iter on socket.

Deprecations:
- crate::payload (use crate::Payload instead)
- crate::error (use crate::Error instead)
- crate::event (use crate::Event instead)


1c3t3a (approved):
Nice!

PR #126, opened 2021-10-04
refactor!: rename socket to client

nshaaban-cPacket:
Breaking changes:
- renamed socket to client. Left a compatibility shim for now if they were using the import in the crate root.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,14 +11,14 @@ An implementation of a socket.io client written in the rust programming language
 ## Example usage
 
 ``` rust
-use rust_socketio::{SocketBuilder, Payload, Socket};
+use rust_socketio::{ClientBuilder, Payload, Client};
 use serde_json::json;
 use std::time::Duration;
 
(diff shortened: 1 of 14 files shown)


nshaaban-cPacket:
> Mhh... what motivates you to do that? Looking at the js examples you always see `socket` [e.g. here](https://socket.io/docs/v4/). But python for example also uses the word `client`.

Socket.IO's client multiplexes over a single network connection using namespaces. Is the underlying connection a socket? Is the client? What about for a server? Is the server itself a socket? Each connection? Each multiplexed interface? The name "Socket" is too vague, and gets confusing. Right now there is a EIO (client) Socket, EIO (shared) Socket, SIO (client) Socket, SIO (shared) Socket. That's not counting the future additional "Socket"s that may be introduced in the server implementation nor the namespace multiplexing, nor the rooms, ect.

1c3t3a:
Oke, that makes sense!

PR #146, opened 2022-01-25
Remove `websocket` dependency and replace it with async bridging code to `tokio-tungstenite`

### LINKED ISSUE #115: Switch websocket library

1c3t3a:
This PR replaces the current websocket code that uses the unmaintained `websocket` crate with [`tokio-tungstenite`](https://github.com/snapview/tokio-tungstenite). `tokio-tungstenite` is async, so in order to get the current code to work I needed to use [async-bridging](https://tokio.rs/tokio/topics/bridging). The idea is that the current websocket transport types hold a `tokio::Runtime` object where they spawn the futures with and then communicate the results like before.

This fixes #115 and does the work of pr #123.

The changes introduce tokio as an async runtime. The current version of the async clients are bound to tokio and it is apparently quite hard to not choose an async runtime, as the underlying connection with the server is currently opened by this crate and thus needs some sort of type (either tokio or async-std streams).

### DIFF
--- a/engineio/src/async_transports/transport.rs
+++ b/engineio/src/async_transports/transport.rs
@@ -6,7 +6,7 @@ use std::time::SystemTime;
 use url::Url;
 
 #[async_trait]
-pub trait AsyncTransport {
+pub(crate) trait AsyncTransport {
     /// Sends a packet to the server. This optionally handles sending of a
     /// socketio binary attachment via the boolean attribute `is_binary_att`.
     async fn emit(&self, data: Bytes, is_binary_att: bool) -> Result<()>;
--- a/engineio/src/async_transports/transport.rs
+++ b/engineio/src/async_transports/transport.rs
@@ -1,0 +1,34 @@
+use crate::error::Result;
+use adler32::adler32;
+use async_trait::async_trait;
+use bytes::Bytes;
+use std::time::SystemTime;
+use url::Url;
+
+#[async_trait]
+pub trait AsyncTransport {
+    /// Sends a packet to the server. This optionally handles sending of a
+    /// socketio binary attachment via the boolean attribute `is_binary_att`.
+    async fn emit(&self, data: Bytes, is_binary_att: bool) -> Result<()>;
+
+    /// Performs the server long polling procedure as long as the client is
+    /// connected. This should run separately at all time to ensure proper
+    /// response handling from the server.
+    async fn poll(&self) -> Result<Bytes>;
+
+    /// Returns start of the url. ex. http://localhost:2998/engine.io/?EIO=4&transport=polling
+    /// Must have EIO and transport already set.
+    async fn base_url(&self) -> Result<Url>;
+
+    /// Used to update the base path, like when adding the sid.
+    async fn set_base_url(&self, base_url: Url) -> Result<()>;
+
+    /// Full query address
+    async fn address(&self) -> Result<Url> {
+        let reader = format!("{:#?}", SystemTime::now());
+        let hash = adler32(reader.as_bytes()).unwrap();
+        let mut url = self.base_url().await?;
+        url.query_pairs_mut().append_pair("t", &hash.to_string());
+        Ok(url)
+    }
+}
--- a/engineio/src/error.rs
+++ b/engineio/src/error.rs
@@ -5,6 +5,7 @@ use std::io::Error as IoError;
 use std::str::Utf8Error;
 use thiserror::Error;
 use url::ParseError as UrlParseError;
+use tungstenite::Error as TungsteniteError;
 use websocket::{client::ParseError, WebSocketError};
 
 /// Enumeration of all possible errors in the `socket.io` context.
@@ -30,6 +31,8 @@ pub enum Error {
     InvalidUrlScheme(String),
     #[error("Error during connection via http: {0}")]
     IncompleteResponseFromReqwest(#[from] ReqwestError),
+    #[error("Error with websocket connection: {0}")]
+    AsyncWebsocketError(#[from] TungsteniteError),
     #[error("Network request returned with status code: {0}")]
     IncompleteHttp(u16),
     #[error("Got illegal handshake response: {0}")]
--- a/engineio/src/error.rs
+++ b/engineio/src/error.rs
@@ -31,7 +31,7 @@ pub enum Error {
     #[error("Error during connection via http: {0}")]
     IncompleteResponseFromReqwest(#[from] ReqwestError),
     #[error("Error with websocket connection: {0}")]
-    AsyncWebsocketError(#[from] TungsteniteError),
+    WebsocketError(#[from] TungsteniteError),
     #[error("Network request returned with status code: {0}")]
     IncompleteHttp(u16),
     #[error("Got illegal handshake response: {0}")]
--- a/engineio/src/error.rs
+++ b/engineio/src/error.rs
@@ -4,9 +4,8 @@ use serde_json::Error as JsonError;
 use std::io::Error as IoError;
 use std::str::Utf8Error;
 use thiserror::Error;
-use url::ParseError as UrlParseError;
 use tungstenite::Error as TungsteniteError;
-use websocket::{client::ParseError, WebSocketError};
+use url::ParseError as UrlParseError;
 
 /// Enumeration of all possible errors in the `socket.io` context.
 #[derive(Error, Debug)]
@@ -39,14 +38,12 @@ pub enum Error {
     InvalidHandshake(String),
     #[error("Called an action before the connection was established")]
     IllegalActionBeforeOpen(),
+    #[error("Error setting up the http request: {0}")]
+    InvalidHttpRequest(#[from] http::Error),
     #[error("string is not json serializable: {0}")]
     InvalidJson(#[from] JsonError),
     #[error("A lock was poisoned")]
     InvalidPoisonedLock(),
-    #[error("Got a websocket error: {0}")]
-    IncompleteResponseFromWebsocket(#[from] WebSocketError),
-    #[error("Error while parsing the url for the websocket connection: {0}")]
-    InvalidWebsocketURL(#[from] ParseError),
     #[error("Got an IO-Error: {0}")]
     IncompleteIo(#[from] IoError),
     #[error("Server did not allow upgrading to websockets")]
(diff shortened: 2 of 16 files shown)


1c3t3a:
Force pushed due to commit rewording.

engineio/src/error.rs
```diff
@@ -30,20 +30,20 @@ pub enum Error {
     InvalidUrlScheme(String),
     #[error("Error during connection via http: {0}")]
     IncompleteResponseFromReqwest(#[from] ReqwestError),
+    #[error("Error with websocket connection: {0}")]
+    AsyncWebsocketError(#[from] TungsteniteError),
```
nshaaban-cPacket, line 34: nit: is the fact that this error came from an async context relevant to users pattern matching on this error?

engineio/src/error.rs
```diff
@@ -30,20 +30,20 @@ pub enum Error {
...
     #[error("Network request returned with status code: {0}")]
     IncompleteHttp(u16),
     #[error("Got illegal handshake response: {0}")]
     InvalidHandshake(String),
     #[error("Called an action before the connection was established")]
     IllegalActionBeforeOpen(),
+    #[error("Error setting up the http request: {0}")]
+    InvalidHttpRequest(#[from] http::Error),
```
nshaaban-cPacket, line 42: nit: Could this also be an invalid response? Or does this type of error only cover requests?

engineio/src/transport.rs
```diff
@@ -1,10 +1,39 @@
...
+    /// connected. This should run separately at all time to ensure proper
+    /// response handling from the server.
+    async fn poll(&self) -> Result<Bytes>;
+
+    /// Returns start of the url. ex. http://localhost:2998/engine.io/?EIO=4&transport=polling
+    /// Must have EIO and transport already set.
+    async fn base_url(&self) -> Result<Url>;
+
+    /// Used to update the base path, like when adding the sid.
+    async fn set_base_url(&self, base_url: Url) -> Result<()>;
```
nshaaban-cPacket, line 25: nit: Dose setting the base_url need to be async, or is it really sync?

engineio/src/transports/websocket.rs
```diff
@@ -1,122 +1,67 @@
...
-        let mut client_builder = WsClientBuilder::new(url[..].as_ref())?;
-        if let Some(headers) = headers {
-            client_builder = client_builder.custom_headers(&headers);
-        }
-        let client = client_builder.connect_insecure()?;
 
-        client.set_nonblocking(false)?;
+        let mut req = http::Request::builder().uri(url.clone().as_str());
+        if let Some(map) = headers {
+            // SAFETY: this unwrap never panics as the underlying request is always in proper state
```
nshaaban-cPacket, line 29: suggestion: Who makes sure it's in the proper state?

engineio/src/transports/websocket.rs
```diff
@@ -1,122 +1,67 @@
...
+use url::Url;
 
 #[derive(Clone)]
 pub struct WebsocketTransport {
-    sender: Arc<Mutex<Writer<TcpStream>>>,
-    receiver: Arc<Mutex<Reader<TcpStream>>>,
-    base_url: Arc<RwLock<url::Url>>,
+    runtime: Arc<Runtime>,
+    inner: Arc<AsyncWebsocketTransport>,
 }
```
nshaaban-cPacket, line 14: No need to do this in this PR but this struct looks more like a generic wrapper around an AsyncTransport implementing Transport

engineio/src/transports_async/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+pub(crate) mod websocket;
+pub(crate) mod websocket_secure;
+
+/// A general purpose asynchronous websocket transport type. Holds
+/// the sender and receiver stream of a websocket connection
+/// and implements the common methods `update`, `poll` and `emit`.
+pub(crate) struct AsyncWebsocketGeneralTransport {
+    sender: Arc<Mutex<SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>>>,
+    receiver: Arc<Mutex<SplitStream<WebSocketStream<MaybeTlsStream<TcpStream>>>>>,
+}
```
nshaaban-cPacket, line 22: Seems a little odd to have this in the mod for all async transports, but I suppose it's shared between websocket and websocket_secure. 

websocket and websocket_secure also appear to be mostly boilerplate

engineio/src/transport.rs
```diff
@@ -1,10 +1,39 @@
 use super::transports::{PollingTransport, WebsocketSecureTransport, WebsocketTransport};
 use crate::error::Result;
 use adler32::adler32;
+use async_trait::async_trait;
 use bytes::Bytes;
 use std::time::SystemTime;
 use url::Url;
 
+#[async_trait]
+pub trait AsyncTransport {
```
nshaaban-cPacket, line 10: nit: Should this be in async_transports?

engineio/src/transports/websocket_secure.rs
```diff
@@ -1,159 +1,77 @@
...
         let mut url = base_url;
         url.query_pairs_mut().append_pair("transport", "websocket");
         url.set_scheme("wss").unwrap();
-        let mut client_builder = WsClientBuilder::new(url[..].as_ref())?;
-        if let Some(headers) = headers {
-            client_builder = client_builder.custom_headers(&headers);
+
+        let mut req = http::Request::builder().uri(url.clone().as_str());
+        if let Some(map) = headers {
+            // SAFETY: this unwrap never panics as the underlying request is always in proper state
```
nshaaban-cPacket, line 35: Who makes sure of that?

engineio/src/transports_async/websocket.rs
```diff
@@ -0,0 +1,67 @@
...
+use url::Url;
+
+use super::AsyncWebsocketGeneralTransport;
+
+/// An asynchronous websocket transport type.
+/// This type only allows for plain websocket
+/// connections ("ws://").
+pub(crate) struct AsyncWebsocketTransport {
+    inner: AsyncWebsocketGeneralTransport,
+    base_url: Arc<RwLock<url::Url>>,
```
nshaaban-cPacket, line 18: nit: use of fully qualified url::Url is not needed after import of url::Url

engineio/src/error.rs
1c3t3a: Good point!

engineio/src/error.rs
1c3t3a: De-facto not, but I see your point...

engineio/src/transport.rs
1c3t3a: It actually needs to be async as I switched to the non-blocking RwLock.

engineio/src/transports/websocket.rs
1c3t3a: Yes it actually is. I am about to build an async version and I think this will be subject to change soon!

engineio/src/transports_async/mod.rs
1c3t3a: They are indeed, the General transport abstracts over those two, the convenient type here is `MaybeTlsStream` allowing for a coexistence. I will take a look on how both of them can be joined...

engineio/src/async_transports/transport.rs
```diff
@@ -0,0 +1,34 @@
+use crate::error::Result;
+use adler32::adler32;
+use async_trait::async_trait;
+use bytes::Bytes;
+use std::time::SystemTime;
+use url::Url;
+
+#[async_trait]
+pub trait AsyncTransport {
```
nshaaban-cPacket, line 9:
```suggestion
pub(crate) trait AsyncTransport {
```

engineio/src/transports/websocket.rs
```diff
@@ -1,122 +1,69 @@
...
 
 impl WebsocketTransport {
     /// Creates an instance of `WebsocketTransport`.
-    pub fn new(base_url: Url, headers: Option<Headers>) -> Result<Self> {
+    pub fn new(base_url: Url, headers: Option<HeaderMap>) -> Result<Self> {
```
nshaaban-cPacket, line 20: suggestion: This is a change in the public interface, will require a semver minor bump.

engineio/src/transports/websocket.rs
```diff
@@ -1,122 +1,69 @@
...
+        let runtime = tokio::runtime::Builder::new_current_thread()
```
nshaaban-cPacket, line 21: nit: do we want it only on one thread on multi threaded?

engineio/src/transports/websocket_secure.rs
```diff
@@ -1,159 +1,77 @@
...
 
 impl WebsocketSecureTransport {
     /// Creates an instance of `WebsocketSecureTransport`.
     pub fn new(
         base_url: Url,
         tls_config: Option<TlsConnector>,
-        headers: Option<Headers>,
+        headers: Option<HeaderMap>,
     ) -> Result<Self> {
+        let runtime = tokio::runtime::Builder::new_current_thread()
```
nshaaban-cPacket, line 25: nit: current thread or multiple?

nshaaban-cPacket (commented):
nit: force-pushing makes it harder to see what changed.

engineio/src/transports/websocket.rs
1c3t3a: Yeah... this is difficult, but the `Headers` type comes from the hyper version that is suspect to a security issue. So using our own type might just be safer... Or we use the [`http::header::HeaderMap`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html) type.

engineio/src/transports/websocket.rs
1c3t3a: True, I see no point in not running this multi-threaded..

engineio/src/transports/websocket.rs
1c3t3a: I decided to use the more popular `http::HeaderMap` and added the proper conversions for it.

nshaaban-cPacket (approved):

1c3t3a:
Thanks a lot for the review @nshaaban-cPacket!

PR #153, opened 2022-02-04
Async engine.io version

1c3t3a:
This PR adds an asynchronous version of the engine.io crate. The async part is hidden behind a feature flag called `async`. The async types involve the following:
 - all three async transports (Polling, WS, WSS).
 - an async `Socket`, that handles the logic for connecting, polling, emitting and calling the corresponding callbacks.
 - an async `OptionalCallback` type, able to hold a `Fn(I) -> impl Future<Output = ()>`.
 - an async `Client` and `ClientBuilder`, that expose the functionality in a set of helpful methods.

Besides tests, this PR also adds a benchmarking suite.

A missing part of this PR is the implementation of [`futures_util::Stream`](https://docs.rs/futures-util/latest/futures_util/stream/trait.Stream.html) as an Iterator for Packets / Bytes. I already started the implementation but struggled with the polling transport as it's blocking during the `poll` request.


engineio/README.md
```diff
@@ -55,3 +55,16 @@ as custom headers for the opening request.
 ## Documentation
 
 Documentation of this crate can be found up on [docs.rs](https://docs.rs/rust_engineio).
+
+## Async version
+
+The crate also ships with an asynchronous version that can be enabled with a feature flag.
+The async version implements the same features mentioned above.
+The asynchronous version has a similar API, just with async functions. Currently the futures
+can only be executed with [`tokio`](https://tokio.rs). In the first benchmarks the async version
```
nshaaban-cPacket, line 64: nit: ideally we wouldn't be tied to tokio. What are we doing that's dependent on tokio?

engineio/benches/engineio.rs
```diff
@@ -45,6 +40,9 @@ pub mod util {
     }
 }
 
+/*            sync benches            */
+
+#[cfg(not(feature = "async"))]
```
nshaaban-cPacket, line 45: suggestion: why are tests turned off but the structs remain on when async is enabled?

engineio/src/asynchronous/async_socket.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    /// Opens the connection to a specified server. The first Pong packet is sent
+    /// to the server to trigger the Ping-cycle.
+    pub async fn connect(&self) -> Result<()> {
+        // SAFETY: Has valid handshake due to type
+        self.connected.store(true, Ordering::Release);
+
+        if let Some(on_open) = self.on_open.as_ref() {
+            let on_open = on_open.clone();
+            self.handle.spawn(async move { on_open(()).await });
```
nshaaban-cPacket, line 74:
```suggestion
            on_open(()).await;
```

nshaaban-cPacket (commented):
Unsure a feature flag is best here, may make more sense to make a separate (but related) async crate (similar to tungstenite) and separate (but related) common crate for traits/structs for packet parsing

engineio/src/asynchronous/client/async_client.rs
```diff
@@ -0,0 +1,629 @@
...
+            let _ = builder(url.clone())
+                .tls_config(
+                    TlsConnector::builder()
+                        .danger_accept_invalid_certs(true)
+                        .build()
+                        .unwrap(),
+                )
+                .build()
+                .await?;
+            let _ = builder(url).headers(headers).build().await?;
```
nshaaban-cPacket, line 625: nit: why suppress must_use?

engineio/README.md
1c3t3a: Basically tokio-tungstenite ;) There is an `async-tungstenite` crate, but it needs an existing stream to abstract over. Might be something for the future, but for now I would settle with tokio. This crate is not capable of running in a `#[no_td]` environment, so being generic for use with the `smol` executor doesn't make sense.

engineio/benches/engineio.rs
1c3t3a: As far as I know the structs are not included, the mod.rs file contains `cfg(feature ...)` attributes.

engineio/src/asynchronous/async_socket.rs
1c3t3a: As far as I know, `await` runs the future to completion before continuing to the next `await` point or the end, that means the method wouldn't exit if the callback hangs or is very slow... The spawn method just spawns a new future in the executor and runs it to completion. The result is `()`, so we don't care about it anyways.

engineio/benches/engineio.rs
```diff
@@ -190,6 +190,236 @@ mod criterion_wrappers {
...
+                    .await
+                    .unwrap()
+                    .close()
+                    .await
+            })
+        });
+    }
+
+    pub fn criterion_engine_io_packet(c: &mut Criterion) {
+        c.bench_function("engine io packet", move |b| b.iter(|| engine_io_packet()));
```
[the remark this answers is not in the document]
1c3t3a: Ahh good point!

engineio/src/asynchronous/async_socket.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::{
+    asynchronous::{callback::OptionalCallback, transport::AsyncTransportType},
+    error::Result,
+    packet::{HandshakePacket, Payload},
+    Error, Packet, PacketId,
+};
+
+#[derive(Clone)]
+pub struct Socket {
+    handle: Handle,
```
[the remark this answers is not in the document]
1c3t3a: See comment above, I am not sure about that.

engineio/src/asynchronous/async_transports/polling.rs
```diff
@@ -0,0 +1,148 @@
...
+
+#[cfg(test)]
+mod test {
+    use tokio::runtime::Builder;
+
+    use crate::asynchronous::transport::AsyncTransport;
+
+    use super::*;
+    use std::str::FromStr;
+    #[test]
```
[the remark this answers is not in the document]
1c3t3a: Good call, I just saw this, didn't know about it.

engineio/src/asynchronous/client/async_client.rs
```diff
@@ -0,0 +1,629 @@
...
+            _ => Err(Error::InvalidUrlScheme(url.scheme().to_string())),
+        }
+    }
+
+    /// Build websocket if allowed, if not allowed or errored fall back to polling.
+    /// WARNING: websocket errors suppressed, no indication of websocket success or failure.
+    pub async fn build_with_fallback(self) -> Result<Client> {
+        let result = self.clone().build().await;
+        if result.is_err() {
+            self.build_polling().await
```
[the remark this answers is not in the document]
1c3t3a: Sure? I don't know about logging in a library, and the doc explains carefully what to expect when calling this method.

engineio/src/asynchronous/client/async_client.rs
```diff
@@ -0,0 +1,629 @@
...
+            // Long lived socket to receive pings
+            let url = crate::test::engine_io_server()?;
+            let socket = builder(url).build().await?;
+
+            socket.connect().await?;
+
+            // hello client
+            let _ = socket.poll().await?;
+            // Ping
+            let _ = socket.poll().await?;
```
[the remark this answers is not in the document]
1c3t3a: Good call!

engineio/README.md
nshaaban-cPacket: Alright, ideally eventually we won't require tokio. It's concerning how many crates depend on tokio being the runtime explicitly.

engineio/src/asynchronous/async_socket.rs
nshaaban-cPacket: Correct. Doesn't really matter either way. If we wanted to remove the handle parameter this change would be needed, otherwise it's fine as is.

engineio/README.md
1c3t3a: True, but it's quite difficult to not be tied to an executor, especially for a networking library. I would go with tokio for now with the aim of changing this in the future.

engineio/src/asynchronous/async_socket.rs
1c3t3a: I don't know about the costs for a `Handle::current()` call. My initial thought was that if it was expensive it might be worth doing it just once and save the handle, but if inexpensive the alternative of just calling `Handle::current().spawn(...)` is more elegant.

nshaaban-cPacket: Either way

1c3t3a: Let's stick with the handle option for now ;)

1c3t3a:
Amended after logical error in test.

engineio/benches/engineio.rs
```diff
@@ -45,6 +40,9 @@ pub mod util {
     }
 }
 
+/*            sync benches            */
```
nshaaban-cPacket, line 43:
```suggestion
/// sync benches
```

engineio/benches/engineio.rs
```diff
@@ -190,6 +190,226 @@ mod criterion_wrappers {
     }
 }
 
+/*            async benches            */
+
```
nshaaban-cPacket, line 194:
```suggestion
/// async benches
```

engineio/src/asynchronous/async_socket.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    // Check if the underlying transport client is connected.
+    pub(crate) fn is_connected(&self) -> Result<bool> {
+        Ok(self.connected.load(Ordering::Acquire))
+    }
+
+    pub(crate) async fn pinged(&self) -> Result<()> {
+        *self.last_ping.lock().await = Instant::now();
+        Ok(())
+    }
```
nshaaban-cPacket, line 176: nit: doesn't need to be Result wrapped.

engineio/src/asynchronous/client/async_client.rs
```diff
@@ -0,0 +1,335 @@
...
+                    // this is already checked during the handshake, so just do nothing here
+                }
+                PacketId::Ping => {
+                    self.socket.pinged().await?;
+                    self.emit(Packet::new(PacketId::Pong, Bytes::new())).await?;
+                }
+                PacketId::Pong => {
+                    // this will never happen as the pong packet is
+                    // only sent by the client
+                    unreachable!();
```
nshaaban-cPacket, line 60: nit: May be better to throw an error here rather than [panicing](https://doc.rust-lang.org/src/core/macros/mod.rs.html#594

nshaaban-cPacket (approved):

1c3t3a:
Great thanks for the review!

> Unsure a feature flag is best here, may make more sense to make a separate (but related) async crate (similar to tungstenite) and separate (but related) common crate for traits/structs for packet parsing

I thought about this and decided to postpone the decision until I see what changes are necessary for the socket.io part. Generally things could get complicated in case we decide to do the same for the socket.io part. In this case we need 4 crates in total.

PR #154, opened 2022-02-11
Implement Stream for all engine.io async types and remove poll methods

1c3t3a:
This PR tries to implement `futures_util::Stream` for all async engine.io types and removes the use of the poll method.

The code still runs into issues with the polling transport. It blocks when executing the polling tests and hangs at this method call: 

https://github.com/1c3t3a/rust-socketio/blob/3f7199cbcbf004bced1f3009909c883dbe402fac/engineio/src/asynchronous/async_transports/polling.rs#L60

The Websocket implementations work fine.

### DIFF
--- a/engineio/src/asynchronous/async_socket.rs
+++ b/engineio/src/asynchronous/async_socket.rs
@@ -1,12 +1,15 @@
 use std::{
     fmt::Debug,
+    pin::Pin,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
     },
+    task::Poll,
 };
 
 use bytes::Bytes;
+use futures_util::{ready, Future, Stream};
 use tokio::{
     runtime::Handle,
     sync::{Mutex, RwLock},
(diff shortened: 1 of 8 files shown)


engineio/src/asynchronous/async_socket.rs
```diff
@@ -23,7 +26,7 @@ use crate::{
 #[derive(Clone)]
 pub struct Socket {
     handle: Handle,
-    transport: Arc<AsyncTransportType>,
+    transport: Arc<Mutex<AsyncTransportType>>,
```
nshaaban-cPacket, line 29: suggestion: a mutex lock shouldn't be needed here, I managed (with some difficultly) to remove it before.

nshaaban-cPacket:
Consider implementing stream on the base transports (Polling, WebSocket, WebSocketSecure) properly (rather just being a wrapper around poll), should simplify the implementation considerably. The websocket types already impl Stream.

1c3t3a:
Closed in favor of #157

PR #155, opened 2022-02-12
Fix dependency line

serg06:


1c3t3a (approved):
Good catch! Thanks a lot ;)

PR #157, opened 2022-02-26
Implement Stream for all engine.io async types and remove poll methods, version 2

1c3t3a:
This is an alternative implementation to #154, that does not implement stream for every type but rather provides a method `stream` on every type that returns a stream.
The stream is mapped on every layer (from acquiring the bytes to parsing packets).
~~The issue with this implementation is the return type of the `Client`s stream method, as this returns a stream of futures which is verbose to handle~~ nvm, I found the `then` method that allows an async context.

I also sticked to reqwest, as tls handling in hyper is quite complicated and reqwest delivers a clean API for configuring a client.

### DIFF
--- a/engineio/src/asynchronous/async_socket.rs
+++ b/engineio/src/asynchronous/async_socket.rs
@@ -23,7 +23,7 @@ use crate::{
 
 #[derive(Clone)]
 pub struct Socket {
-    pub(super) handle: Handle,
+    handle: Handle,
     transport: Arc<AsyncTransportType>,
     on_close: OptionalCallback<()>,
     on_data: OptionalCallback<Bytes>,
--- a/engineio/src/asynchronous/async_socket.rs
+++ b/engineio/src/asynchronous/async_socket.rs
@@ -1,5 +1,6 @@
 use std::{
     fmt::Debug,
+    pin::Pin,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
@@ -7,11 +8,11 @@ use std::{
 };
 
 use bytes::Bytes;
-use tokio::{
-    runtime::Handle,
-    sync::{Mutex, RwLock},
-    time::Instant,
+use futures_util::{
+    stream::{self, once},
+    Stream, StreamExt,
 };
+use tokio::{runtime::Handle, sync::Mutex, time::Instant};
 
 use crate::{
     asynchronous::{callback::OptionalCallback, transport::AsyncTransportType},
@@ -22,7 +23,7 @@ use crate::{
 
 #[derive(Clone)]
 pub struct Socket {
-    handle: Handle,
+    pub(super) handle: Handle,
     transport: Arc<AsyncTransportType>,
     on_close: OptionalCallback<()>,
     on_data: OptionalCallback<Bytes>,
@@ -33,8 +34,6 @@ pub struct Socket {
     last_ping: Arc<Mutex<Instant>>,
     last_pong: Arc<Mutex<Instant>>,
     connection_data: Arc<HandshakePacket>,
-    /// Since we get packets in payloads it's possible to have a state where only some of the packets have been consumed.
-    remaining_packets: Arc<RwLock<Option<crate::packet::IntoIter>>>,
 }
 
 impl Socket {
@@ -59,7 +58,6 @@ impl Socket {
             last_ping: Arc::new(Mutex::new(Instant::now())),
             last_pong: Arc::new(Mutex::new(Instant::now())),
             connection_data: Arc::new(handshake),
-            remaining_packets: Arc::new(RwLock::new(None)),
         }
     }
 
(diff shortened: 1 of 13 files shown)


engineio/src/asynchronous/async_socket.rs
```diff
@@ -83,37 +81,45 @@ impl Socket {
...
-                    let iter = iter.as_mut().unwrap();
-                    if let Some(packet) = iter.next() {
-                        return Ok(Some(packet));
-                    }
-                }
+    /// Creates a stream over the incoming packets, uses the streams provided by the
+    /// underlying transport types.
+    pub(crate) async fn stream(
+        &self,
+    ) -> Result<Pin<Box<dyn Stream<Item = Result<Option<Packet>>> + '_>>> {
```
nshaaban-cPacket, line 88: suggestion: return `impl Stream<Item = Result<Option<Packet>>>`

engineio/src/asynchronous/async_transports/polling.rs
```diff
@@ -80,14 +81,21 @@ impl AsyncTransport for PollingTransport {
...
 
-    async fn poll(&self) -> Result<Bytes> {
-        Ok(self
-            .client
-            .get(self.address().await?)
-            .send()
-            .await?
-            .bytes()
-            .await?)
+    async fn stream(&self) -> Result<Pin<Box<dyn Stream<Item = Result<Bytes>> + '_>>> {
```
nshaaban-cPacket, line 84: suggestion: return impl Stream

engineio/src/asynchronous/async_transports/websocket_general.rs
```diff
@@ -89,3 +100,17 @@ impl AsyncWebsocketGeneralTransport {
...
+    fn poll_next(
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        match ready!(Pin::new(&mut Box::pin(self.inner.poll())).poll(cx)) {
+            Ok(val) => Poll::Ready(Some(Ok(val))),
+            Err(err) => Poll::Ready(Some(Err(err))),
+        }
+    }
+}
```
nshaaban-cPacket, line 116: nit: why? The websocket crate impls Stream

engineio/src/asynchronous/async_socket.rs
```diff
@@ -83,37 +81,45 @@ impl Socket {
...
-                    // SAFETY: checked is some above
-                    let mut iter = self.remaining_packets.write().await;
-                    let iter = iter.as_mut().unwrap();
-                    if let Some(packet) = iter.next() {
-                        return Ok(Some(packet));
-                    }
-                }
+    /// Creates a stream over the incoming packets, uses the streams provided by the
+    /// underlying transport types.
+    pub(crate) async fn stream(
```
nshaaban-cPacket, line 86: suggestion: does this need to be async?

engineio/src/asynchronous/client/async_client.rs
```diff
@@ -29,44 +32,48 @@ impl Client {
...
-                }
-                PacketId::Message => {
-                    self.socket.handle_data(packet.data.clone()).await;
-                }
-                PacketId::Close => {
-                    self.socket.handle_close().await;
-                }
-                PacketId::Open => {
-                    unreachable!("Won't happen as we open the connection beforehand");
+    pub async fn stream(&self) -> Result<Pin<Box<dyn Stream<Item = Result<Option<Packet>>> + '_>>> {
```
nshaaban-cPacket, line 36: suggestion: impl stream

engineio/src/asynchronous/async_transports/polling.rs
```diff
@@ -80,14 +81,21 @@ impl AsyncTransport for PollingTransport {
...
+        let address = self.address().await?;
```
nshaaban-cPacket, line 85: suggestion: turn this into a stream, then map it to be a stream of streams then flatten rather than awaiting

engineio/src/asynchronous/async_transports/polling.rs
1c3t3a: This is apparently not possible, as `impl "Trait"` is not allowed in a trait definition at the moment...

engineio/src/asynchronous/async_socket.rs
1c3t3a: At the moment this is async as the address of a transport is locked with an async lock. If I switch to sync locking this will be fine, but blocking...

engineio/src/asynchronous/client/async_client.rs
1c3t3a: Same as for the other transports.

engineio/src/asynchronous/async_transports/polling.rs
1c3t3a: Good catch!!

engineio/src/asynchronous/async_transports/polling.rs
nshaaban-cPacket: OH right, forgot about that.

engineio/src/asynchronous/async_transports/websocket_general.rs
```diff
@@ -80,37 +73,22 @@ impl AsyncWebsocketGeneralTransport {
...
-        cx: &mut std::task::Context<'_>,
-    ) -> std::task::Poll<Option<Self::Item>> {
-        match ready!(Pin::new(&mut Box::pin(self.inner.poll())).poll(cx)) {
-            Ok(val) => Poll::Ready(Some(Ok(val))),
-            Err(err) => Poll::Ready(Some(Err(err))),
-        }
+    pub(crate) fn stream(&self) -> impl Stream<Item = Result<Bytes>> + '_ {
+        self.receiver
+            .lock()
+            .into_stream()
```
nshaaban-cPacket, line 79: nit: does this into_stream consume self.receiver?

nshaaban-cPacket (approved):

engineio/src/asynchronous/async_transports/websocket_general.rs
1c3t3a: I don't think so, as far as I understand this consumes the future, otherwise tests like this: https://github.com/1c3t3a/rust-socketio/blob/db405b1661c4be7d779868b3374914a2e3740e47/engineio/src/asynchronous/async_transports/websocket.rs#L144-L155
wouldn't work as the two calls to `stream` each resolve in a different stream.

1c3t3a: Or am I wrong about this?

nshaaban-cPacket: Maybe a failing of the naming of the other crate. If it works, it works.

PR #165, opened 2022-03-12
Implement socketio auth

marekvospel:
Add option to send data in socketio opening request, to authenticate with the server.
https://socket.io/docs/v4/client-options/#auth

### DIFF
--- a/ci/socket-io-auth.js
+++ b/ci/socket-io-auth.js
@@ -1,0 +1,18 @@
+const server = require('http').createServer();
+const io = require('socket.io')(server);
+
+console.log('Started');
+var callback = client => {
+    console.log('Connected!');;
+    client.on('test', function (arg, ack) {
+        console.log('Ack received')
+        if (ack) {
+            const payload = client.handshake.auth.password === '123' ? '456' : '789'
+            ack(payload);
+        }
+    });
+};
+io.on('connection', callback);
+io.of('/admin').on('connection', callback);
+// the socket.io client runs on port 4201
+server.listen(4204);
--- a/ci/socket-io-auth.js
+++ b/ci/socket-io-auth.js
@@ -14,5 +14,5 @@ var callback = client => {
 };
 io.on('connection', callback);
 io.of('/admin').on('connection', callback);
-// the socket.io client runs on port 4201
+// the socket.io client runs on port 4204
 server.listen(4204);
--- a/socketio/src/client/builder.rs
+++ b/socketio/src/client/builder.rs
@@ -184,14 +184,13 @@ impl ClientBuilder {
     ///     .namespace("/admin")
     ///     .auth(json!({ "password": "1337" }))
     ///     .on("error", |err, _| eprintln!("Error: {:#?}", err))
-    ///     .connect();
+    ///     .connect()
+    ///     .expect("Connection error");
     ///
     /// ```
-    pub fn auth<T: TryInto<serde_json::Value>>(mut self, auth: T) -> Self {
-        self.auth = match auth.try_into() {
-            Ok(value) => Some(value),
-            Err(_) => None,
-        };
+    pub fn auth<T: Into<serde_json::Value>>(mut self, auth: T) -> Self {
+        self.auth = Some(auth.into());
+
         self
     }
 
--- a/socketio/src/client/builder.rs
+++ b/socketio/src/client/builder.rs
@@ -36,6 +36,7 @@ pub struct ClientBuilder {
     tls_config: Option<TlsConnector>,
     opening_headers: Option<HeaderMap>,
     transport_type: TransportType,
+    auth: Option<String>,
 }
 
 impl ClientBuilder {
@@ -77,6 +78,7 @@ impl ClientBuilder {
             tls_config: None,
             opening_headers: None,
             transport_type: TransportType::Any,
+            auth: None,
         }
     }
 
@@ -172,6 +174,13 @@ impl ClientBuilder {
         self
     }
 
+    //TODO: add documentation
+    pub fn auth<T: Into<String>>(mut self, auth: T) -> Self {
+        self.auth = Some(auth.into());
+
+        self
+    }
+
     /// Specifies which EngineIO [`TransportType`] to use.
     /// # Example
     /// ```rust
@@ -261,7 +270,7 @@ impl ClientBuilder {
 
         let inner_socket = InnerSocket::new(engine_client)?;
 
-        let socket = Client::new(inner_socket, &self.namespace, self.on)?;
+        let socket = Client::new(inner_socket, &self.namespace, self.on, self.auth)?;
         socket.connect()?;
 
         Ok(socket)
--- a/socketio/src/client/builder.rs
+++ b/socketio/src/client/builder.rs
@@ -174,7 +174,19 @@ impl ClientBuilder {
         self
     }
 
-    //TODO: add documentation
+    /// Sets data sent in the opening request.
+    /// # Example
+    /// ```rust
+    /// use rust_socketio::{ClientBuilder};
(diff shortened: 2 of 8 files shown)


socketio/src/client/client.rs
```diff
@@ -524,6 +535,37 @@ mod test {
...
+        let (tx, rx) = mpsc::sync_channel(0);
+
+        assert!(socket
+            .emit_with_ack(
+                "test",
+                json!({ "msg": "1" }),
+                Duration::from_secs(1),
+                move |payload, _| {
+                    println!("Got ack");
+                    tx.send(Payload::String(json!(["456"]).to_string()) == payload)
```
1c3t3a, line 555: You could make use of the impl of `From<Value>` for Payload.
```diff
- tx.send(Payload::String(json!(["456"]).to_string()) == payload)
+ tx.send(Payload::from(json!(["456"])) == payload)
```

socketio/src/client/client.rs
```diff
@@ -524,6 +535,37 @@ mod test {
...
+                        .unwrap();
+                }
+            )
+            .is_ok());
+
+        sleep(Duration::from_secs(2));
+
+        let received = rx.recv();
```
1c3t3a, line 563: The `rx.recv()` call blocks, why do we need to sleep beforehand?

socketio/src/client/builder.rs
```diff
@@ -36,6 +36,7 @@ pub struct ClientBuilder {
     tls_config: Option<TlsConnector>,
     opening_headers: Option<HeaderMap>,
     transport_type: TransportType,
+    auth: Option<String>,
```
1c3t3a, line 39: I think it might be better to use `serve_json::Value` here instead of string. In that case we're sure that the provided string is valid Json.

socketio/src/client/builder.rs
```diff
@@ -172,6 +174,25 @@ impl ClientBuilder {
...
+    /// use serde_json::json;
+    ///
+    /// let socket = ClientBuilder::new("http://localhost:4204/")
+    ///     .namespace("/admin")
+    ///     .auth(json!({ "password": "1337" }).to_string())
+    ///     .on("error", |err, _| eprintln!("Error: {:#?}", err))
+    ///     .connect();
+    ///
+    /// ```
+    pub fn auth<T: Into<String>>(mut self, auth: T) -> Self {
```
1c3t3a, line 190: If using `serde_json::Value`, it makes sense to change the bound to `Into<serde_json::Value>` or even `TryInto<serde_json::Value>` if we want to allow a `&str` as input.

socketio/src/client/client.rs
```diff
@@ -46,12 +48,14 @@ impl Client {
         socket: InnerSocket,
         namespace: T,
         on: HashMap<Event, Callback>,
+        auth: Option<String>,
```
1c3t3a, line 51: Same as above (`serde_json::Value`).

socketio/src/lib.rs
```diff
@@ -124,4 +124,19 @@ pub(crate) mod test {
 
         url
     }
+
+    // The socket.io auth server for testing runs on port 4204
+    const AUTH_SERVER_URL: &str = "http://localhost:4204";
+
+    pub(crate) fn socket_io_auth_server() -> Url {
+        let url =
+            std::env::var("SOCKET_IO_AUTH_SERVER").unwrap_or_else(|_| AUTH_SERVER_URL.to_owned());
```
1c3t3a, line 133: Could you please also add the `SOCKET_IO_AUTH_SERVER` env variable here: https://github.com/1c3t3a/rust-socketio/blob/3dce3e1e690847676a2e465adc49d0483e181aed/.devcontainer/docker-compose.yaml#L45

ci/socket-io-auth.js
```diff
@@ -0,0 +1,18 @@
...
+        console.log('Ack received')
+        if (ack) {
+            const payload = client.handshake.auth.password === '123' ? '456' : '789'
+            ack(payload);
+        }
+    });
+};
+io.on('connection', callback);
+io.of('/admin').on('connection', callback);
+// the socket.io client runs on port 4201
```
1c3t3a, line 17:
```diff
- // the socket.io client runs on port 4201
+ // the socket.io client runs on port 4204
```

1c3t3a (changes requested):
Thanks for the PR! I never actually noticed that we're missing the auth part here, so thanks for implementing it ;) I added some thoughts and comments to your code. If you need logs on why the test failed in CI, you can download the Artifacts from `Actions/<<your commit's generate coverage Action>>/Artifacts` (or [this link](https://github.com/1c3t3a/rust-socketio/actions/runs/1973046815) for the current commit). Please let me know if you don't have access to this part of an action (I don't really know about the permissions TBH ;) ).

marekvospel:
Hey, thanks for the quick code review, I've changed everything you requested and the CI should now also pass. :)

socketio/src/client/client.rs
marekvospel: I just forgot to delete that line :D

marekvospel:
The test running that long is caused by the emit with ack being received before the listener in the socket io server is registered.
image
I'll add a 1s sleep, so the server has time to register all listeners.

socketio/src/client/builder.rs
```diff
@@ -172,6 +174,27 @@ impl ClientBuilder {
...
+    ///     .namespace("/admin")
+    ///     .auth(json!({ "password": "1337" }))
+    ///     .on("error", |err, _| eprintln!("Error: {:#?}", err))
+    ///     .connect();
+    ///
+    /// ```
+    pub fn auth<T: TryInto<serde_json::Value>>(mut self, auth: T) -> Self {
+        self.auth = match auth.try_into() {
+            Ok(value) => Some(value),
+            Err(_) => None,
```
1c3t3a, line 193: In this case we are discarding the error and don't tell the user about it, which is not good (imagine a situation where you need to debug the client and realize this is the bug :D). I think it's better to return a ´Result<Self>` here and map the error to our own socket.io error (´Error::InvalidJson` respectively).

1c3t3a (changes requested):
Great, thanks for the quick reaction! Just a nit and were good to go

socketio/src/client/builder.rs
1c3t3a: Ahhh my bad! `serde_json::Value` does not provide a `Try_From<String>` implementation. In that case `T: Into<serde_json::Value>` would be the best. Sorry for the confusion!

socketio/src/client/client.rs
```diff
@@ -62,8 +66,12 @@ impl Client {
         // Connect the underlying socket
         self.socket.connect()?;
 
+        let auth = self.auth.as_ref().map(|data| data.to_string());
+
+        println!("{:?}", auth);
```
1c3t3a, line 71: This shouldn't be here ;)

socketio/src/client/client.rs
```diff
@@ -524,6 +532,36 @@ mod test {
...
+    fn socket_io_auth_builder_integration() -> Result<()> {
+        let url = crate::test::socket_io_auth_server();
+        let socket = ClientBuilder::new(url)
+            .namespace("/admin")
+            .auth(json!({ "password": "123" }))
+            .connect()?;
+
+        let (tx, rx) = mpsc::sync_channel(0);
+
+        sleep(Duration::from_secs(1));
```
1c3t3a, line 545: Maybe add a comment on why this is present, just for the future.

1c3t3a (changes requested):
Alright, just the println and we're fine

1c3t3a (approved):
LGTM! Thanks ;)

PR #168, opened 2022-03-16
refactor: change `stream` implementation in engine.io

1c3t3a:
This PR changes the `stream` implementation and usage in the engine.io trait. Earlier PRs introduced async Streams in engine.io but the ergonomics were quite difficult and the implementation was hard to understand. This is changed in this PR, with the introduction of [`async_stream`](https://docs.rs/async-stream/0.3.3/async_stream/index.html), a crate that exposes a handy `try_stream!` macro for writing async generators that implement `futures_util::Stream`.

The changes made here leave some open questions on how sound the generator implementation actually is. Especially for the long polling case, the following issue arises: 
https://github.com/1c3t3a/rust-socketio/blob/a643468d2bf5b7e6db870264ce62e1c6b526377f/engineio/src/asynchronous/async_transports/polling.rs#L96-L119

A solution would be to use something like [`stream_cancel`](https://docs.rs/stream-cancel/0.8.1/stream_cancel/index.html) which wraps a stream and makes it cancelable from outside the stream. 

From my understanding the websocket generator should be sound. It uses a `Mutex` and each generator (spawned by a call to `client.stream()`) works on the same websocket client. The websocket stream is done as soon as the connection to the server is properly closed. An open issue might be that the locking implementation with the `Mutex` might lead to unperformant polling in case many generators want to access the same data.

Until this issue is resolved, I am marking this PR as a Draft PR.

### DIFF
--- a/engineio/src/asynchronous/async_socket.rs
+++ b/engineio/src/asynchronous/async_socket.rs
@@ -97,9 +97,6 @@ impl Socket {
             PacketId::Close => {
                 self.handle_close();
             }
-            PacketId::Open => {
-                unreachable!("Won't happen as we open the connection beforehand");
-            }
             PacketId::Upgrade => {
                 // this is already checked during the handshake, so just do nothing here
             }
@@ -107,9 +104,9 @@ impl Socket {
                 self.pinged().await;
                 self.emit(Packet::new(PacketId::Pong, Bytes::new())).await?;
             }
-            PacketId::Pong => {
-                // this will never happen as the pong packet is
-                // only sent by the client
+            PacketId::Pong | PacketId::Open => {
+                // this will never happen as the pong and open
+                // packets are only sent by the client
                 return Err(Error::InvalidPacket());
             }
             PacketId::Noop => (),
--- a/engineio/src/asynchronous/async_socket.rs
+++ b/engineio/src/asynchronous/async_socket.rs
@@ -129,7 +129,7 @@ impl Socket {
     /// underlying transport types.
     fn stream(
         mut transport: AsyncTransportType,
-    ) -> Pin<Box<impl Stream<Item = Result<Packet>> + 'static>> {
+    ) -> Pin<Box<impl Stream<Item = Result<Packet>> + 'static + Send>> {
         // map the byte stream of the underlying transport
         // to a packet stream
         Box::pin(try_stream! {
@@ -182,14 +182,10 @@ impl Socket {
             packet.into()
         };
 
-        if let Err(error) = self
-            .transport
-            .lock()
-            .await
-            .as_transport()
-            .emit(data, is_binary)
-            .await
-        {
+        let lock = self.transport.lock().await;
+        let fut = lock.as_transport().emit(data, is_binary);
+
+        if let Err(error) = fut.await {
             self.call_error_callback(error.to_string());
             return Err(error);
         }
--- a/engineio/src/asynchronous/async_socket.rs
+++ b/engineio/src/asynchronous/async_socket.rs
@@ -6,11 +6,9 @@ use std::{
     },
 };
 
+use async_stream::try_stream;
 use bytes::Bytes;
-use futures_util::{
-    stream::{self, once},
-    Stream, StreamExt,
-};
+use futures_util::Stream;
 use tokio::{runtime::Handle, sync::Mutex, time::Instant};
 
 use crate::{
@@ -80,38 +78,33 @@ impl Socket {
         Ok(())
     }
 
+    /// Helper method that parses bytes and returns an iterator over the elements.
+    fn parse_payload(bytes: Bytes) -> impl Stream<Item = Result<Packet>> {
+        try_stream! {
+            let payload = Payload::try_from(bytes);
+
+            for elem in payload?.into_iter() {
+                yield elem;
+            }
+        }
+    }
+
     /// Creates a stream over the incoming packets, uses the streams provided by the
     /// underlying transport types.
-    pub(crate) fn stream(&self) -> Result<impl Stream<Item = Result<Option<Packet>>> + '_> {
+    pub(crate) fn stream(&self) -> Result<impl Stream<Item = Result<Packet>> + '_> {
         if !self.connected.load(Ordering::Acquire) {
             return Err(Error::IllegalActionBeforeOpen());
         }
 
         // map the byte stream of the underlying transport
         // to a packet stream
-        let stream = self.transport.as_transport().stream()?.map(move |payload| {
-            // this generator is needed as Rust only allows one right_stream call per closure
-            let error_gen = |err| once(async { Err(err) }).right_stream();
-
-            if let Err(err) = payload {
-                return error_gen(err);
+        Ok(Box::pin(try_stream! {
+            for await payload in self.transport.as_transport().stream() {
+                for await packet in Self::parse_payload(payload?) {
+                    yield packet?;
+                }
             }
-
-            // SAFETY: checked is some above
-            let payload = Payload::try_from(payload.unwrap());
-            if let Err(err) = payload {
-                return error_gen(err);
-            }
-
-            // SAFETY: checked is some above
-            stream::iter(payload.unwrap().into_iter())
-                .map(|val| Ok(Some(val)))
-                .left_stream()
-        });
-
-        // Flatten stream of streams into stream
-        let stream = stream.flatten();
-        Ok(Box::pin(stream))
+        }))
     }
 
     pub async fn disconnect(&self) -> Result<()> {
--- a/engineio/src/asynchronous/async_socket.rs
+++ b/engineio/src/asynchronous/async_socket.rs
@@ -1,5 +1,6 @@
 use std::{
     fmt::Debug,
+    pin::Pin,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
@@ -8,7 +9,7 @@ use std::{
 
 use async_stream::try_stream;
 use bytes::Bytes;
-use futures_util::Stream;
+use futures_util::{ready, FutureExt, Stream, StreamExt};
 use tokio::{runtime::Handle, sync::Mutex, time::Instant};
 
 use crate::{
@@ -18,10 +19,12 @@ use crate::{
     Error, Packet, PacketId,
 };
 
+use super::generator::Generator;
+
 #[derive(Clone)]
 pub struct Socket {
     handle: Handle,
-    transport: Arc<AsyncTransportType>,
+    transport: Arc<Mutex<AsyncTransportType>>,
     on_close: OptionalCallback<()>,
     on_data: OptionalCallback<Bytes>,
     on_error: OptionalCallback<String>,
@@ -31,6 +34,7 @@ pub struct Socket {
     last_ping: Arc<Mutex<Instant>>,
     last_pong: Arc<Mutex<Instant>>,
     connection_data: Arc<HandshakePacket>,
+    generator: Arc<Mutex<Generator<Result<Packet>>>>,
 }
 
 impl Socket {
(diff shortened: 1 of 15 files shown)


engineio/src/asynchronous/async_socket.rs
```diff
@@ -80,38 +78,33 @@ impl Socket {
         Ok(())
     }
 
+    /// Helper method that parses bytes and returns an iterator over the elements.
```
nshaaban-cPacket, line 81: oooo very nice!

engineio/src/asynchronous/async_transports/polling.rs
```diff
@@ -81,31 +93,29 @@ impl AsyncTransport for PollingTransport {
...
-                Err(err) => async { Err(err) }.right_future(),
-            })
-            .then(|resp| async {
-                match resp.await {
-                    Ok(val) => val
-                        .bytes_stream()
-                        .map_err(Error::IncompleteResponseFromReqwest)
-                        .left_stream(),
-                    Err(err) => once(async { Err(err) }).right_stream(),
+    fn stream(&self) -> Pin<Box<dyn Stream<Item = Result<Bytes>> + '_>> {
```
nshaaban-cPacket, line 96: suggestion: consume self and return impl Stream
```suggestion
    fn stream(self) -> impl Stream<Item = Result<Bytes>> + '_> {
```

A transport should be converted into a stream, then no longer referenced as a transport.

We may need to setup some channels in order to allow for sending messages

engineio/src/asynchronous/async_transports/polling.rs
```diff
@@ -81,31 +93,29 @@ impl AsyncTransport for PollingTransport {
...
+
+            // TODO: Investigate a way to stop this. From my understanding this generator
+            // lives on the heap (as it's boxed) and always sends long polling
+            // requests. In the current version of the crate it is possible to create a stream
+            // using an `&self` of client, which would mean that we potentially end up with multiple
+            // of these generators on the heap. That is not what we want.
+            // A potential solution would be to wrap the client (also helpful for websockets) and make
+            // sure to call the "poll" request only on the same object over and over again. In that
+            // case the underlying object would have information about whether the connection is shut down
+            // or not.
```
nshaaban-cPacket, line 108: Consuming self should fix this issue

engineio/src/asynchronous/async_transports/websocket.rs
```diff
@@ -151,8 +151,8 @@ mod test {
                 transport.base_url().await?.to_string()
             )
         );
-        println!("{:?}", transport.stream()?.next().await.unwrap());
-        println!("{:?}", transport.stream()?.next().await.unwrap());
+        println!("{:?}", transport.stream().next().await.unwrap());
+        println!("{:?}", transport.stream().next().await.unwrap());
```
nshaaban-cPacket, line 155: nit: shouldn't this be a comparison (to properly test) rather than a print?

engineio/src/asynchronous/async_transports/websocket_general.rs
```diff
@@ -73,22 +74,29 @@ impl AsyncWebsocketGeneralTransport {
...
 
+    fn receive_message(&self) -> impl Stream<Item = Result<Message>> + '_ {
+        try_stream! {
+            for await item in &mut *self.receiver.lock().await {
+                yield item?;
+            }
+        }
+    }
+
     pub(crate) fn stream(&self) -> impl Stream<Item = Result<Bytes>> + '_ {
```
nshaaban-cPacket, line 85: suggestion: consume self, should make Mutex uneeded.

engineio/src/asynchronous/async_transports/polling.rs
1c3t3a: Yes, that's a good idea. We only need to separate sender and receiver in that case, as we still want to call methods on the sender after turning the receiver into a stream. I will look into that.

nshaaban-cPacket: May be worth looking into how the Websocket crate implements .split (as that's effectively what we want to do). 

Either that or setup message channels

1c3t3a: Yeah that comes with an implementation of `Sink` and `Stream` for a type at the same time... https://docs.rs/futures-util/latest/futures_util/stream/trait.StreamExt.html#method.split

1c3t3a: I've looked into this, implementing Stream is... difficult - to put it nicely.... I think it is a lot of code to get it running for all transports.

nshaaban-cPacket: Could you impl stream by just wrapping an existing stream?
```rust
struct IImplStreamSinkSplit {
stream: Box<dyn Stream<Item=Bytes>>
sink: Box<dyn Sink<Item=Bytes>>
}
struct PollingTransport {}

impl PollingTransport {
  fn into_stream_sink(self) -> (impl Stream<Item=Bytes>, impl Sink<Item=Bytes>) {
    let stream = try_stream! {};
    let sink = try_sink!{}; // Not sure how sinks work
    }
}

impl From<PollingTransport> for ImplStreamSink {
  fn from(polling: obj) -> Self {
    let (stream, sink) = self.into_stream_sink()
    Self {
      stream: Box::new(stream),
      sink: Box::new(sink)
    }
  }
}


```

1c3t3a: I implemented this but end up with a deadlock... the thing gets messy as the generator always need mutable access to the sender and receiver. Locking the client in the `try_stream!` macro results in a deadlock... Uff this is tough..

1c3t3a: Yes I think the current version would be fine if we only call the `stream` method once. However I solved the deadlock issue and I am going to commit the new version now. If it's fine with you we could discuss the advantages / disadvantages of both versions ;)

1c3t3a:
I am going to rebase after a review of the current commit. This got a bit messy as I rebased on main in between....

engineio/src/asynchronous/async_socket.rs
```diff
@@ -80,38 +83,74 @@ impl Socket {
...
+                self.handle_data(packet.data.clone());
+            }
+            PacketId::Message => {
+                self.handle_data(packet.data.clone());
+            }
+            PacketId::Close => {
+                self.handle_close();
+            }
+            PacketId::Open => {
+                unreachable!("Won't happen as we open the connection beforehand");
```
nshaaban-cPacket, line 101: nit: consider throwing InvalidPacket. if there is a bug in the lib here, panicking would be annoying for library consumers (ideally libraries never panic, just error)

nshaaban-cPacket (approved):

engineio/src/asynchronous/async_socket.rs
1c3t3a: Yeah good point! ;)

socketio/src/client/client.rs
```diff
@@ -304,7 +304,7 @@ impl Client {
     #[inline]
     fn handle_binary_event(&self, packet: &Packet) -> Result<()> {
         let event = if let Some(string_data) = &packet.data {
-            string_data.replace("\"", "").into()
+            string_data.replace('\"', " ").into()
```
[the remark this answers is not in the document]
1c3t3a: https://rust-lang.github.io/rust-clippy/master/#single_char_pattern

engineio/src/asynchronous/async_transports/websocket_general.rs
```diff
@@ -72,26 +75,41 @@ impl AsyncWebsocketGeneralTransport {
...
-                }
+                            msg.freeze()
+                        } else if msg.is_text() {
+                            Bytes::from(msg.into_data())
+                        } else {
+                            // in the edge case of a packet with a type other than text
+                            // or binary map it to an empty packet.
+                            // this always needs to get filtered out
+                            // by the user of the stream
+                            Bytes::new()
```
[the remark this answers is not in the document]
1c3t3a: Yes.... that's a bit difficult as I could only map the stream in that situation (we're in the stream implementation here...)

engineio/src/asynchronous/client/async_client.rs
```diff
@@ -147,23 +148,27 @@ mod test {
     }
 
     async fn test_connection(socket: Client) -> Result<()> {
```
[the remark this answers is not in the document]
1c3t3a: Ah this message is called from other test scenarios with a specific socket and can't be seen as a freestanding test.

engineio/src/asynchronous/async_transports/websocket_general.rs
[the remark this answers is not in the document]
1c3t3a: Oke I found a way

socketio/src/client/client.rs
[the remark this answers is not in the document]
nshaaban-cPacket: Why is this replace here? Why did it get switched to a space?

engineio/src/asynchronous/async_transports/websocket_general.rs
```diff
@@ -86,30 +86,25 @@ impl Stream for AsyncWebsocketGeneralTransport {
...
-                            // or binary map it to an empty packet.
-                            // this always needs to get filtered out
-                            // by the user of the stream
-                            Bytes::new()
-                        }
-                    })
-                })
-            })
-            .map_err(Error::WebsocketError)
+        loop {
```
nshaaban-cPacket, line 89: suggestion: why loop? this match always returns.

engineio/src/asynchronous/async_transports/websocket_general.rs
```diff
@@ -86,30 +86,25 @@ impl Stream for AsyncWebsocketGeneralTransport {
...
+            let next = ready!(lock.poll_next_unpin(cx));
+
+            match next {
```
nshaaban-cPacket, line 92: nit: all of these return Poll::Ready, consider
```
let result = match next {
...
};

Poll::Ready(result)
```

socketio/src/client/client.rs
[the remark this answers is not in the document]
1c3t3a: Ahh that was a mistake! I didn't mean to insert the space here! ;)

engineio/src/asynchronous/async_transports/websocket_general.rs
1c3t3a: I refactored the loop in the newest commit. Generally the idea is to poll for a new packet (skip the current one) in case the packet has a type other than message or binary.

nshaaban-cPacket: If the type is other than text/binary wouldn't that be a Poll::Pending? See https://github.com/1c3t3a/rust-socketio/pull/168/files/31d3712f2021f738dc9daa386ddd1b50be8d2330..7dd1cd1d0b55094e71b058951e1e7a8e4f78fdba#r837707732

engineio/src/asynchronous/async_transports/websocket_general.rs
```diff
@@ -84,24 +84,21 @@ impl Stream for AsyncWebsocketGeneralTransport {
...
+                Some(Ok(Message::Text(str))) => return Poll::Ready(Some(Ok(Bytes::from(str)))),
+                Some(Ok(Message::Binary(data))) => {
+                    let mut msg = BytesMut::with_capacity(data.len() + 1);
+                    msg.put_u8(PacketId::Message as u8);
+                    msg.put(data.as_ref());
+
+                    return Poll::Ready(Some(Ok(msg.freeze())));
                 }
+                // ignore packets other than text and binary
+                Some(Ok(_)) => (),
```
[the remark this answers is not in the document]
1c3t3a: Ahh you're right, good call! ;)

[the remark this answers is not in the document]
1c3t3a: Oke apparently `Poll::Pending` doesn't work accordingly. When I try to receive a message *after* I ignore the packet (with returning `Pending`), the new frame is not processed and the underlying websocket transport never yields a new message.... To test this scenario I opened a connection to a server that waited for 30s before sending a socketio message, the client maintained an open connection for over 40s. After 25s, the client received the servers WS pong message and returned `Pending`. The next message that was send after 30s was never received. The current loop based approach handled this correctly.

nshaaban-cPacket (approved):

1c3t3a:
Force pushed in order to get the history straight

PR #169, opened 2022-03-17
Implement socketio onAny

marekvospel:
Add option to listen on any recived event
https://socket.io/docs/v4/client-api/#socketonanycallback
Update socket_io_auth_builder_integration (recive "auth" event instead of emitting with ack)

### DIFF
--- a/socketio/src/client/callback.rs
+++ b/socketio/src/client/callback.rs
@@ -4,21 +4,25 @@ use std::{
 };
 
 use super::Client;
-use crate::Payload;
+use crate::{Event, Payload};
 
-type InnerCallback = Box<dyn for<'a> FnMut(Payload, Client) + 'static + Sync + Send>;
+pub(crate) type SocketCallback = Box<dyn for<'a> FnMut(Payload, Client) + 'static + Sync + Send>;
+pub(crate) type SocketAnyCallback =
+    Box<dyn for<'a> FnMut(Event, Payload, Client) + 'static + Sync + Send>;
 
-pub(crate) struct Callback {
-    inner: InnerCallback,
+pub(crate) struct Callback<T> {
+    inner: T,
 }
 
-impl Debug for Callback {
+// SocketCallback implementations
+
+impl Debug for Callback<SocketCallback> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.write_str("Callback")
     }
 }
 
-impl Deref for Callback {
+impl Deref for Callback<SocketCallback> {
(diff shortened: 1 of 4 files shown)


socketio/src/client/callback.rs
```diff
@@ -4,35 +4,39 @@ use std::{
...
-type InnerCallback = Box<dyn for<'a> FnMut(Payload, Client) + 'static + Sync + Send>;
+pub(crate) type SocketCallback = Box<dyn for<'a> FnMut(Payload, Client) + 'static + Sync + Send>;
+pub(crate) type SocketAnyCallback =
+    Box<dyn for<'a> FnMut(Event, Payload, Client) + 'static + Sync + Send>;
 
-pub(crate) struct Callback {
-    inner: InnerCallback,
+pub(crate) struct Callback<T> {
+    inner: T,
 }
```
1c3t3a, line 15: I don't know if being generic here is the right way.... It's not to be expected that we have any other kind of callback. In that case an enum is usually the way to go, but that is difficult with the deref impl, maybe we find a way around? Otherwise we could just have to callback structs.

socketio/src/client/client.rs
```diff
@@ -238,9 +242,22 @@ impl Client {
...
+        }
+        match event.clone() {
+            Event::Message | Event::Custom(_) => {
+                if let Some(callback) = on_any_lock {
+                    callback(event.clone(), payload, self.clone())
+                }
+            }
+            _ => {}
         }
         drop(on);
```
1c3t3a, line 262: Either add a `drop(on_any)` here or remove this line, the lock is dropped on the next call anyways as it's residing on the stack.

socketio/src/client/client.rs
```diff
@@ -530,33 +546,59 @@ mod test {
...
+            .on("auth", |payload, _client| {
+                if let Payload::String(msg) = payload {
+                    println!("{}", msg)
+                }
+            })
+            .on_any(|event, payload, _client| {
+                if let Payload::String(str) = payload {
+                    println!("{} {}", String::from(event), str);
+                }
+            })
```
1c3t3a, line 564: Could we somehow assure that this is called with multiple events?

1c3t3a (changes requested):
Thanks for tackling the issue!

socketio/src/client/callback.rs
marekvospel: I tought socket.io server implementation would also use different callbacks.

1c3t3a: Yeah that is true, but probarbly of the same type. Anyhow, lets use this for a second that way.

1c3t3a (approved):

1c3t3a:
Thanks!

PR #170, opened 2022-03-22
Add a GitHub action that runs an engineio benchmark

1c3t3a:
I want to give this a try. It might be the case that this doesn't run reliable in a GitHub action environment. If that is the case we could remove the action again.


1c3t3a:
@benchmark

1c3t3a:
@benchmark

PR #210, opened 2022-08-05
chore(socketio::client): callback match event no need clone

SSebo:


1c3t3a (approved):
Good catch! Thank you 👍🏽

PR #212, opened 2022-08-10
chore(socket): use match instead of unwrap

SSebo:


socketio/src/socket.rs
```diff
@@ -159,12 +159,11 @@ impl Socket {
...
-        if let Err(err) = socket_packet {
-            return Err(err);
-        }
-        // SAFETY: checked above to see if it was Err
-        let mut socket_packet = socket_packet.unwrap();
+        let mut socket_packet = match Packet::try_from(&packet.data) {
+            Err(err) => return Err(err),
+            Ok(packet) => packet,
+        };
+
```
1c3t3a, line 166: This can in fact be written as:
```rust
 let socket_packet = Packet::try_from(&packet.data)?;
```
so no need for a match here.

1c3t3a (changes requested):
Thanks for the PR! I would appreciate if you could concatenate all cleanup you do into one PR, that is easier to review for me!

PR #213, opened 2022-08-10
fix(websocket): lock conflict between emit and poll

SSebo:


1c3t3a:
Hi @SSebo, first of all thanks for the contribution and for finding this bug! That is great :) For your PR I am not so sure about the implementation. The async transport never actually locks at the same time, so the only problem is the sync transport. A minimal inversiv and nice change would be to find a way for removing the lock from the sync transport. I assume this should be possible as all methods from the Transport interface only take `&self` instead of .`&mut self`. Your solution would work, but basically gives access to another clone of the sender which is harder to read and more difficult to reason about.

SSebo:
@1c3t3a the code is more reasonable now, PTAL

1c3t3a (approved):
LGTM, thanks a lot!

PR #217, opened 2022-08-12
Reconnect

SSebo:
Still under construction

- [ ] Reconnect

- [ ] Retry

- [ ] Backoff


SSebo:
hi, @1c3t3a could you help to figure out why the coverage github action can not pass, but the test github action is ok ?

SSebo:
now coverage github action passed, wired.

SSebo:
hi @1c3t3a, the restart testing case will make other case failed, so I add a new `socket-io-restart.js` to do restart operation. #225 PTAL

SSebo:
@1c3t3a Could you help me to figure out why the test pass but coverage action failed?

socketio/src/client/client.rs
```diff
@@ -468,6 +568,53 @@ mod test {
...
 
+    #[test]
+    fn socket_io_reconnect_integration() -> Result<()> {
+        let url = crate::test::socket_io_restart_server();
+
+        let socket = ClientBuilder::new(url).connect();
+        assert!(socket.is_ok(), "should connect success");
+        let socket = socket.unwrap();
+
+        std::thread::sleep(std::time::Duration::from_millis(200));
```
nshaaban-cPacket, line 579: nit: why sleep?

socketio/src/client/builder.rs
```diff
@@ -302,24 +284,31 @@ impl ClientBuilder {
...
-            TransportType::Websocket => builder.build_websocket()?,
-            TransportType::WebsocketUpgrade => builder.build_websocket_with_upgrade()?,
-        };
-
-        let inner_socket = InnerSocket::new(engine_client)?;
-
+        let transport_type = self.transport_type.clone();
         let socket = Client::new(
-            inner_socket,
+            Box::new(move || build_socket(transport_type.clone(), builder.clone())),
```
nshaaban-cPacket, line 289: (creative vision difference) code smell: crossing concerns, seems very odd to pass a copy of the builder in a function into the underlying client it's trying to build

socketio/src/client/client.rs
```diff
@@ -468,6 +568,53 @@ mod test {
...
+        assert!(r.is_ok(), "should emit restart success");
+
+        let mut is_emit_error = false;
+        for _ in 0..10 {
+            std::thread::sleep(std::time::Duration::from_millis(500));
+            let r = socket.emit("message", json!(""));
+            if r.is_err() {
+                is_emit_error = true; // server is down
+                break;
+            }
```
nshaaban-cPacket, line 594: (usability question) nit: should the re-connection be completely transparent to end user? As a user if I had to attempt to repeat all of my emits that may make it hard to utilize this feature

socketio/src/client/client.rs
```diff
@@ -468,6 +568,53 @@ mod test {
...
+        }
+
+        assert_eq!(
+            is_emit_error, true,
+            "should emit failed because server is down"
+        );
+
+        for _ in 0..10 {
```
nshaaban-cPacket, line 602: suggestion: why is this looped? Should the next message right after the failure return success?

socketio/src/client/client.rs
```diff
@@ -585,143 +735,115 @@ mod test {
...
-            packet,
-            Packet::new(
-                PacketId::Event,
-                nsp.clone(),
-                Some("[\"auth\",\"success\"]".to_owned()),
-                None,
-                0,
-                None
-            )
-        );
```
nshaaban-cPacket, line 613: suggestion: The iterator SHOULD be tested by the unit tests, the message pipe is clever, but more complicated that needed for these tests. 

It's two separate concerns:
- testing behavior (via the iterator)
- callback interface is functioning 

If the callback code breaks, it's much more difficult to debug than the raw iterator implementation.

socketio/src/client/client.rs
```diff
@@ -585,143 +735,115 @@ mod test {
...
-        let packet = packet.unwrap();
-
-        assert_eq!(
-            packet,
-            Packet::new(
-                PacketId::Event,
-                nsp.clone(),
-                Some("[\"Hello from the message event!\"]".to_owned()),
-                None,
-                0,
```
nshaaban-cPacket, line 672: suggestion: Is using the Iterator interface on Client no longer functional?

nshaaban-cPacket (commented):
Like to start off by saying this is a valid implementation, as well as effort went into adding tests, which is much appreciated. 

Concerned about the additional complication (addition of arc/mutex/rwlocks) in the underlying client, as well as increased potential in IllegalActionBeforeOpen edge cases. Neither of which are currently present on master.

Perhaps a wrapper around client would be cleaner separating the auto-reconnect logic from the normal client logic? Auto re-connection seems more of a concern for a wrapper around client rather than complicating the underlying client implementation. 

At the end of the day this is a valid implementation, and for how it's done, it looks good (besides is one usability question). While I may prefer a different approach, decision is up to @1c3t3a . Feel free to merge without my approval.

@SSebo Thanks for the PR, sorry for not documenting purpose of the connect_manual, nor intention of iterators use in the future. I recommenced waiting for @1c3t3a before attempting to implement my comments related to differing creative vision (I've annotated the comments in question), as what you have is valid and functional.

Thank you for properly implementing unit tests for your features and expanding the existing test infrastructure.

socketio/src/client/client.rs
SSebo: Good catch, I found connect function return not means connect success, the first emit message will not success if not wait.

socketio/src/client/client.rs
```diff
@@ -394,6 +452,47 @@ impl Client {
...
+        std::thread::spawn(move || {
+            // tries to restart a poll cycle whenever a 'normal' error occurs,
+            // it just panics on network errors, in case the poll cycle returned
+            // `Result::Ok`, the server receives a close frame so it's safe to
+            // terminate
+            for packet in self_clone.iter() {
+                if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = packet {
+                    //TODO: 0.3.X handle errors
+                    self_clone.reconnect();
+                    panic!("{}", e.unwrap_err());
```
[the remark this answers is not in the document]
SSebo: The panic is the old logic, I agree with you to just log.

socketio/src/client/builder.rs
SSebo: @nshaaban-cPacket,  I agree it is odd. But It is hard to implement reconnecting if under laying socket creation is part of builder. user code hold the socket client, the builder is gone after `build()`, all the info about how to create socket is not available. Would you give some good suggestion?

socketio/src/client/client.rs
SSebo: @nshaaban-cPacket Server not start immediately, emit will not success until server is start. If I change the test server to restart immediately, user code may emit success every time, we have no idea if the server is restart or not.

socketio/src/client/client.rs
nshaaban-cPacket: blocker: That issue doesn't exist in main, connect should block until the client is usable.

socketio/src/client/builder.rs
nshaaban-cPacket: Agreed you'll need the context of the builder in order to create a new client. Consider 
> Perhaps a wrapper around client would be cleaner separating the auto-reconnect logic from the normal client logic? Auto re-connection seems more of a concern for a wrapper around client rather than complicating the underlying client implementation.

socketio/src/client/client.rs
nshaaban-cPacket: (usability issue) seems difficult for users of the library to use

SSebo:
> Like to start off by saying this is a valid implementation, as well as effort went into adding tests, which is much appreciated.

I agree with you, It is not a very clean implementation. I would like to refactor this PR.

> Concerned about the additional complication (addition of arc/mutex/rwlocks) in the underlying client, as well as increased potential in IllegalActionBeforeOpen edge cases. Neither of which are currently present on master.

Yes, but if reconnecting feature is needed, the failed socket must be replaced with new one, so there must be an arc lock. otherwise provide a notification, to let user create a new one. As you said we can try not transparent reconnect.

> Perhaps a wrapper around client would be cleaner separating the auto-reconnect logic from the normal client logic? Auto re-connection seems more of a concern for a wrapper around client rather than complicating the underlying client implementation.

Good idea.

> At the end of the day this is a valid implementation, and for how it's done, it looks good (besides is one usability question). While I may prefer a different approach, decision is up to @1c3t3a . Feel free to merge without my approval.

I would like to refactor the code, your approval is important.

> @SSebo Thanks for the PR, sorry for not documenting purpose of the connect_manual, nor intention of iterators use in the future. I recommenced waiting for @1c3t3a before attempting to implement my comments related to differing creative vision (I've annotated the comments in question), as what you have is valid and functional.
> 
> Thank you for properly implementing unit tests for your features and expanding the existing test infrastructure.

I am new here, so much to learn.

nshaaban-cPacket:
Wonderful! I'm almost done with a PoC that should get most of the way with the wrapper I was talking about.

> Yes, but if reconnecting feature is needed, the failed socket must be replaced with new one, so there must be an arc lock. otherwise provide a notification, to let user create a new one. As you said we can try not transparent reconnect.

Yes there must be some Arc/RwLocks somewhere to get around the callbacks not being Clone, but there is some flexibility as to where ;)

1c3t3a:
I agree with @nshaaban-cPacket here. the proposed new solution in #229 looks better! Anyways thanks a lot for going the extra mile @SSebo on reacting to the comments here and thanks a lot @nshaaban-cPacket for your expertise and the example PR! The wrapper sounds like the perfect solution, especially since we are wrapping types in this library on each level anyways, so I think this fits the bigger picture ;)

PR #218, opened 2022-08-14
fix: decode utf8

SSebo:
hi @1c3t3a,  the original decode logic treat payload as vec[u8] which is ok for unicode string, but not ok for utf8 which is `variable-width`.


1c3t3a (approved):
LGTM!

Issue #221, opened 2022-08-14
On close callback not called when server down


SSebo:
if the server down without send close packet, `rust_socketio` can not call `.on(Event::Close, |_, _| {})` callback.

the `socket.io-client` will do that

PR #223, opened 2022-08-15
fix: on close callback when server down without close packet

SSebo:
#221

### DIFF
--- a/socketio/src/client/client.rs
+++ b/socketio/src/client/client.rs
@@ -143,9 +143,10 @@ impl Client {
         let disconnect_packet =
             Packet::new(PacketId::Disconnect, self.nsp.clone(), None, None, 0, None);
 
-        self.socket.send(disconnect_packet)?;
+        let _ = self.socket.send(disconnect_packet);
         self.socket.disconnect()?;
 
+        let _ = self.callback(&Event::Close, ""); // trigger on_close
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


socketio/src/client/client.rs
```diff
@@ -143,9 +143,10 @@ impl Client {
         let disconnect_packet =
             Packet::new(PacketId::Disconnect, self.nsp.clone(), None, None, 0, None);
 
-        self.socket.send(disconnect_packet)?;
+        let _ = self.socket.send(disconnect_packet);
```
nshaaban-cPacket, line 146: This would be a good location for logging, not sure if there's a plan for logging yet in this crate @1c3t3a

1c3t3a: There is no plan yet, but I am happy for ideas around it. For know this would exceed the scope of this PR, but maybe add a comment to remember at a later point?

nshaaban-cPacket: IMO https://docs.rs/tracing/latest/tracing/ at (mostly) debug level (NOT INFO). Like how https://github.com/hyperium/hyper/blob/5ec094caa5c999e6f919a2bc82f5f3b7d40c2d8a/src/proto/h2/client.rs does it.

nshaaban-cPacket (approved):
Would be nice to add a `// TODO` for later logging

SSebo:
@1c3t3a @nshaaban-cPacket `TODO` added. PTAL

1c3t3a (approved):
👍🏽  Thanks ;)

PR #229, opened 2022-08-19
feat: reconnect refactor

SSebo:
@nshaaban-cPacket This PR is based on your POC 👍 
@nshaaban-cPacket @1c3t3a PTAL


engineio/src/socket.rs
```diff
@@ -79,7 +79,7 @@ impl Socket {
             spawn_scoped!(on_close(()));
         }
 
-        self.emit(Packet::new(PacketId::Close, Bytes::new()))?;
+        let _ = self.emit(Packet::new(PacketId::Close, Bytes::new())); // will not succeed when server is gone
```
1c3t3a, line 82: nit: Please place the comment on line 81 and not after the code ;)

socketio/examples/callback.rs
```diff
@@ -35,9 +35,8 @@ fn main() {
         // Event specified by enum
         .on(Event::Close, |_, socket| {
             println!("Socket Closed");
-            socket
-                .emit("message", json!({"foo": "Hello server"}))
-                .expect("Error emitting");
+            let r = socket.emit("message", json!({"foo": "Hello server"}));
```
1c3t3a, line 38: nit: please name the variable `err` or `error`.

socketio/src/client/builder.rs
```diff
@@ -30,15 +31,20 @@ pub enum TransportType {
...
-    on_any: Option<Callback<SocketAnyCallback>>,
+    on: Arc<RwLock<HashMap<Event, Callback<SocketCallback>>>>,
+    on_any: Arc<RwLock<Option<Callback<SocketAnyCallback>>>>,
     namespace: String,
     tls_config: Option<TlsConnector>,
     opening_headers: Option<HeaderMap>,
     transport_type: TransportType,
     auth: Option<serde_json::Value>,
+    pub reconnect: bool,
+    pub reconnect_attempts: u8, // 0 means infinity
```
1c3t3a, line 45: nit: Please place the comment on line 44.

socketio/src/client/builder.rs
```diff
@@ -75,13 +81,17 @@ impl ClientBuilder {
...
-            on_any: None,
+            on: Arc::new(RwLock::new(HashMap::new())),
+            on_any: Arc::new(RwLock::new(None)),
             namespace: "/".to_owned(),
             tls_config: None,
             opening_headers: None,
             transport_type: TransportType::Any,
             auth: None,
+            reconnect: true,
+            reconnect_attempts: 0, // infinity
```
1c3t3a, line 92: nit: Please place the comment on line 91.

socketio/src/client/builder.rs
```diff
@@ -96,6 +106,24 @@ impl ClientBuilder {
...
+    }
+
+    pub fn reconnect_delay(mut self, min: u64, max: u64) -> Self {
+        self.reconnect_delay = min;
+        self.reconnect_delay_max = max;
+
+        self
+    }
+
+    // reconnect_attempts 0 means infinity
```
1c3t3a, line 121: nit: Rephrase to: `Zero reconnect attempts represent infinity`.

socketio/src/client/client.rs
```diff
@@ -238,10 +238,17 @@ impl Client {
...
     pub(crate) fn iter(&self) -> Iter {
         Iter { socket: self }
     }
 
+    // for reconnect test
+    #[allow(dead_code)]
+    pub(crate) fn is_connected(&self) -> Result<bool> {
+        self.socket.is_engineio_connected()
+    }
+
```
1c3t3a, line 251: nit: Please rename to `is_connected_testing`.

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,216 @@
+#![allow(unused)]
+use std::{
+    ops::Deref,
+    sync::{Arc, Mutex, RwLock},
+    time::Duration,
+};
+
+pub use super::super::{event::Event, payload::Payload};
```
1c3t3a, line 8: nit: Is it better to use the absolute path here? `super::super` looks a bit odd, but I might be wrong here and this is common practice.

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,216 @@
...
+use super::{Client, ClientBuilder};
+use crate::{error::Result, packet::Packet, Error};
+use backoff::ExponentialBackoff;
+use backoff::{backoff::Backoff, ExponentialBackoffBuilder};
+
+#[derive(Clone)]
+pub struct ReconnectClient {
+    builder: ClientBuilder,
+    client: Arc<RwLock<super::Client>>,
```
1c3t3a, line 17: nit: why is the super needed here?

socketio/src/client/builder.rs
1c3t3a, line 45: This is the maximum amount of allowed reconnects, the variable name suggests it is a counter, wdyt about renaming the variable to something like `max_reconnect_attempts`?

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl ReconnectClient {
+    pub(crate) fn new(builder: ClientBuilder) -> Result<Self> {
+        let builder_clone = builder.clone();
+        let client = builder_clone.connect_manual()?;
+        let backoff = ExponentialBackoffBuilder::new()
+            .with_initial_interval(Duration::from_millis(builder.reconnect_delay))
+            .with_max_interval(Duration::from_millis(builder.reconnect_delay_max))
+            .build();
+        let max_reconnect_attempts = builder.reconnect_attempts;
```
1c3t3a, line 30: Why do we need this variable here? Can't we just ise the builder in the Self construction?

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,216 @@
...
+        timeout: Duration,
+        callback: F,
+    ) -> Result<()>
+    where
+        F: for<'a> FnMut(Payload, Client) + 'static + Sync + Send,
+        E: Into<Event>,
+        D: Into<Payload>,
+    {
+        let client = self.client.read()?;
+        // TODO: like js client, buffer emit, resend after reconnect
```
1c3t3a, line 66: nit: Can you please create an issue for that and then write `TODO(#issue-nr)`?

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub fn disconnect(&self) -> Result<()> {
+        let client = self.client.read()?;
+        client.disconnect()
+    }
+
+    fn reconnect(&mut self) -> Result<()> {
+        let mut reconnect_attempts = 0;
+        if self.builder.reconnect {
+            loop {
+                if self.max_reconnect_attempts != 0 // 0 means infinity
```
1c3t3a, line 79: nit: Please move comment to line 78.

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,216 @@
...
+                }
+                reconnect_attempts += 1;
+
+                if let Some(backoff) = self.backoff.next_backoff() {
+                    std::thread::sleep(backoff);
+                }
+
+                if self.do_reconnect().is_err() {
+                    self.do_reconnect();
+                }
```
1c3t3a, line 92: I don't really understand why we need to do this again if it fails the first time, can you elaborate?

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,216 @@
...
+        // Use thread to consume items in iterator in order to call callbacks
+        std::thread::spawn(move || {
+            // tries to restart a poll cycle whenever a 'normal' error occurs,
+            // it just panics on network errors, in case the poll cycle returned
+            // `Result::Ok`, the server receives a close frame so it's safe to
+            // terminate
+            for packet in self_clone.iter() {
+                if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = packet {
+                    //TODO: 0.3.X handle errors
+                    println!("{}", e.unwrap_err());
```
1c3t3a, line 123: Please don't log here (yet!). We can add a TODO that this is a place where logging makes sense.

1c3t3a (changes requested):
This looks pretty good, only nits and a few minor things! Thanks for going the extra mile and adding such thorough tests, that is really good to have. The wrapper is a much cleaner solution as the one proposed in #217, thanks a lot for reacting to the feedback!

socketio/src/client/reconnect.rs
SSebo: The right one should be this:
```rust 
if self.do_reconnect().is_ok() {
      break;
}
```
I'll fix it.

socketio/src/client/reconnect.rs
SSebo: You are right!

socketio/src/client/reconnect.rs
SSebo: I agree, this is IDE auto added. I'll fix it.

socketio/src/client/reconnect.rs
SSebo: Good catch!

socketio/src/client/builder.rs
SSebo: I borrowed from the js client, but I think max_reconnect_attempts is better

SSebo:
I found the test case is not stable. I'll try to fix it.

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,239 @@
...
+        // TODO(#230): like js client, buffer emit, resend after reconnect
+        client.emit_with_ack(event, data, timeout, callback)
+    }
+
+    pub fn disconnect(&self) -> Result<()> {
+        let client = self.client.read()?;
+        client.disconnect()
+    }
+
+    fn reconnect(&mut self) -> Result<()> {
```
nshaaban-cPacket, line 72:
```suggestion
    fn reconnect(&mut self) {
```

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,239 @@
...
+            // tries to restart a poll cycle whenever a 'normal' error occurs,
+            // it just panics on network errors, in case the poll cycle returned
+            // `Result::Ok`, the server receives a close frame so it's safe to
+            // terminate
+            for packet in self_clone.iter() {
+                if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = packet {
+                    //TODO: 0.3.X handle errors
+                    //TODO: logging error
+                    let _ = self_clone.disconnect();
+                    break;
```
nshaaban-cPacket, line 124:
```suggestion
                    //TODO: logging error
                    let _ = self_clone.disconnect();
                    let self_clone.reconnect()
```

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,239 @@
...
+                }
+            }
+            self_clone.reconnect();
```
nshaaban-cPacket, line 127: iterator has the rwlock, we can continue to call next on the iterator.
```suggestion
                    //TODO: logging error
                    let _ = self_clone.disconnect();
                    self_clone.reconnect()
                }
            }
```

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,239 @@
...
+            .reconnect(true)
+            .max_reconnect_attempts(0) // infinity
+            .reconnect_delay(100, 100)
+            .on(Event::Connect, move |_, socket| {
+                let mut count = connect_count_clone.lock().unwrap();
+                *count += 1;
+            })
+            .on(Event::Close, move |_, socket| {
+                let mut count = close_count_clone.lock().unwrap();
+                *count += 1;
```
nshaaban-cPacket, line 186: Did you mean -1? Or not including this at all? Is Close even called/
```suggestion
                *count -= 1;
```

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,239 @@
...
+            assert_eq!(
+                *connect_cnt, 2,
+                "should connect twice while server is gone and back"
+            );
+            assert_eq!(*close_cnt, 1, "should close once while server is gone");
+        }
+
+        socket.disconnect()?;
+        Ok(())
+    }
```
nshaaban-cPacket, line 238: suggestion: test the iterator implementation and make sure there is a constant stream of packets, even when reconnecting

socketio/src/socket.rs
```diff
@@ -102,12 +102,13 @@ impl Socket {
...
-                } else {
-                    payload = format!("[\"{}\",\"{}\"]", String::from(event), str_data);
-                }
+                let payload = {
+                    if serde_json::from_str::<serde_json::Value>(&str_data).is_ok() {
+                        format!("[\"{}\",{}]", String::from(event), str_data)
+                    } else {
+                        format!("[\"{}\",\"{}\"]", String::from(event), str_data)
+                    }
+                };
```
nshaaban-cPacket, line 111: nit: style 
```suggestion
                let payload = if serde_json::from_str::<serde_json::Value>(&str_data).is_ok() {
                    format!("[\"{}\",{}]", String::from(event), str_data)
                } else {
                    format!("[\"{}\",\"{}\"]", String::from(event), str_data)
                };
```

nshaaban-cPacket (changes requested):
Couple blockers mostly related to maintaining public APIs. If you want to learn highly recommend [this](https://rust-lang.github.io/api-guidelines/about.html) (for some reason hard to find) rust book on maintaining public APIs. The actual implementation looks much cleaner! Thanks!

socketio/src/client/reconnect.rs
SSebo: how about add 
```rust
.on("message", move |_, socket| {
                // test the iterator implementation and make sure there is a constant
                // stream of packets, even when reconnecting
                let mut message_cnt = message_count_clone.lock().unwrap();
                *message_cnt += 1;
            })
```

socketio/src/client/reconnect.rs
SSebo: I checked the js client, `on('disconnet', () => {})` and  `on('close',, () => {})` both will be called when disconnecting, since we just have `Event::Close`, so `.on(Event::Close, || {})` should be called I think.

socketio/src/client/builder.rs
```diff
@@ -263,25 +294,8 @@ impl ClientBuilder {
...
-                if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = packet {
-                    //TODO: 0.3.X handle errors
-                    panic!("{}", e.unwrap_err())
-                }
-            }
-        });
-
-        Ok(socket)
+    pub fn connect(self) -> Result<ReconnectClient> {
+        ReconnectClient::new(self)
```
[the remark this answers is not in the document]
SSebo: The API is compatible (examples work without change). so the user code can compile and use as before. Should we add a new API even though?

socketio/src/client/client.rs
```diff
@@ -238,6 +238,7 @@ impl Client {
         }
     }
 
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
SSebo: there is a lint warning on this, I checked code, only test function used this code.

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,239 @@
...
+        let client = builder_clone.connect_manual()?;
+        let backoff = ExponentialBackoffBuilder::new()
+            .with_initial_interval(Duration::from_millis(builder.reconnect_delay_min))
+            .with_max_interval(Duration::from_millis(builder.reconnect_delay_max))
+            .build();
+
+        let s = Self {
+            builder,
+            client: Arc::new(RwLock::new(client)),
+            backoff: ExponentialBackoff::default(),
```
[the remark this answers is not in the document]
SSebo: Good catch!

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    pub fn disconnect(&self) -> Result<()> {
+        let client = self.client.read()?;
+        client.disconnect()
+    }
+
+    fn reconnect(&mut self) -> Result<()> {
+        let mut reconnect_attempts = 0;
+        if self.builder.reconnect {
```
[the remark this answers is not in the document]
SSebo: I checked js client, they can set `reconnect = false` to disable reconnect. we can leave `connect` API to normal Client, but add a new API like `reconnectable_connect`(not a good name) to get rid of `reconnect: bool` variable.

socketio/src/client/builder.rs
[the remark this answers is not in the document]
nshaaban-cPacket:
```rust
let Client = builder.connect()?;
```
Would have passed before, and would now fail. We'd have to replace Client with ReconnectingClient in the public interface matching the path to make this not breaking.

socketio/src/client/client.rs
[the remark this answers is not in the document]
nshaaban-cPacket: In that case
```suggestion
    #[cfg(test)]
```

socketio/src/client/reconnect.rs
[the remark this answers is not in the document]
nshaaban-cPacket: (another option)
solution three: rename ReconnctingClient to Client (and take it's place), and rename existing client to InternalClient

Thoughts @1c3t3a

socketio/src/client/reconnect.rs
nshaaban-cPacket: Just smells odd that both connect and close should increment the same counter both upwards

socketio/src/client/reconnect.rs
nshaaban-cPacket: Prefer if the iterator itself was tested (as the callbacks are meerly another layer on top of the iterator), but that would help a tiny bit

nshaaban-cPacket (changes requested):
Still unresolved blockers, seems like some of the comments were marked as resolved when they were not resolved

socketio/src/client/reconnect.rs
SSebo: it is not same count，just the lock guard same name，I renamed in new commit

SSebo:
@nshaaban-cPacket looks like not review the latest code?

socketio/src/client/reconnect.rs
```diff
@@ -248,4 +248,56 @@ mod test {
         socket.disconnect()?;
         Ok(())
     }
+
+    #[test]
+    fn socket_io_iterator_integration() -> Result<()> {
```
SSebo, line 253: @nshaaban-cPacket Is this test for iterator OK? ReconnectClient have to be manual created to not call `poll_callback` function.

socketio/src/client/builder.rs
```diff
@@ -90,8 +90,8 @@ impl ClientBuilder {
             transport_type: TransportType::Any,
             auth: None,
             reconnect: true,
-            // 0 means infinity
-            max_reconnect_attempts: 0,
+            // None means infinity
+            max_reconnect_attempts: None,
```
SSebo, line 94: @1c3t3a max_reconnect_attempts is option now

socketio/src/client/reconnect.rs
```diff
@@ -163,74 +159,90 @@ mod test {
...
+                    |_, _| {},
+                );
+                assert!(r.is_ok(), "should emit message success");
             })
-            .on(Event::Close, move |_, socket| {
-                let mut count = close_count_clone.lock().unwrap();
-                *count += 1;
+            .on(Event::Close, move |_, _| {
+                let mut close_cnt = close_count_clone.lock().unwrap();
+                *close_cnt += 1;
```
SSebo, line 192: @nshaaban-cPacket  renamed to `connect_cnt` and `close_cnt`

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,308 @@
...
+impl Iterator for Iter {
+    type Item = Result<Packet>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let socket = self.socket.read();
+        match socket {
+            Ok(socket) => match socket.poll() {
+                Err(err) => Some(Err(err)),
+                Ok(Some(packet)) => Some(Ok(packet)),
+                Ok(None) => Some(Err(Error::PollInterrupted)),
```
SSebo, line 141: @nshaaban-cPacket I found iter will break if inner socket is dropped, so PollInterrupted is added.

nshaaban-cPacket (approved):

socketio/src/client/reconnect.rs
SSebo: How about MissingInnerSocket ? packet may not be missing?

SSebo:
>  solution three: rename ReconnctingClient to Client (and take it's place), and rename existing client to InternalClient

I tried, found that InternalClient is hard to hide from user because of the callback parameter.  Compiler told me that trait is not object safe if I move emit function with generic type in.

socketio/src/client/reconnect.rs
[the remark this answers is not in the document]
1c3t3a: I agree with you @nshaaban-cPacket, this makes sense. Changing the Clients default reconnection behavior is a breaking change anyways, so it would make sense to release a new breaking version and be transparent about what has changed.

engineio/src/socket.rs
```diff
@@ -79,7 +79,8 @@ impl Socket {
             spawn_scoped!(on_close(()));
         }
 
-        self.emit(Packet::new(PacketId::Close, Bytes::new()))?;
+        // will not succeed when server is gone
```
1c3t3a, line 82:
```suggestion
        // will not succeed when connection to the server is interrupted
```

socketio/examples/callback.rs
```diff
@@ -35,9 +35,8 @@ fn main() {
         // Event specified by enum
         .on(Event::Close, |_, socket| {
             println!("Socket Closed");
-            socket
-                .emit("message", json!({"foo": "Hello server"}))
-                .expect("Error emitting");
+            let result = socket.emit("message", json!({"foo": "Hello server"}));
+            assert!(result.is_err());
```
1c3t3a, line 39: Why this change?

1c3t3a (commented):
> > solution three: rename ReconnctingClient to Client (and take it's place), and rename existing client to InternalClient
> 
> @nshaaban-cPacket I tried, found that InternalClient is hard to hide from user because of the callback parameter. Compiler told me that trait is not object safe if I move emit function with generic type in.

Could you share the compiler message? Would be interesting to see why.

socketio/examples/callback.rs
SSebo: `emit().expect()` will panic when `on_close` called which will not reconnect when server is interrupted.

socketio/src/client/reconnect.rs
```diff
@@ -0,0 +1,251 @@
...
+    use native_tls::TlsConnector;
+    use serde_json::json;
+    use std::time::{Duration, Instant};
+
+    #[test]
+    fn socket_io_reconnect_integration() -> Result<()> {
+        let url = crate::test::socket_io_restart_server();
+        let connect_count = Arc::new(Mutex::new(0));
+        let close_count = Arc::new(Mutex::new(0));
+        let message_count = Arc::new(Mutex::new(0));
```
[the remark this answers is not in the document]
SSebo: Yes, cleaner than Arc Mutex

SSebo:
@1c3t3a Compiler messages: 

--> socketio/src/client/reconnect.rs:40:8
    |
14  | pub trait ClientTrait {
    |           ----------- this trait cannot be made into an object...
...
40  |     fn emit<E, D>(&self, event: E, data: D) -> Result<()>
    |        ^^^^ ...because method `emit` has generic type parameters
...
80  |     fn emit_with_ack<F, E, D>(
    |        ^^^^^^^^^^^^^ ...because method `emit_with_ack` has generic type parameters
    = help: consider moving `emit` to another trait
    = help: consider moving `emit_with_ack` to another trait

For more information about this error, try `rustc --explain E0038`.
error: could not compile `rust_socketio` due to previous error

nshaaban-cPacket:
If this is a breaking change anyways, we could expose BOTH ReconnectingClient and Client in the public interface

socketio/examples/callback.rs
1c3t3a: I see, but `assert!` is also resulting in a panic (see source [comment](https://doc.rust-lang.org/src/core/macros/mod.rs.html#1373-1376)). So where is the difference? The error message?

1c3t3a:
> @1c3t3a Compiler does not complain until I create a trait object `Box<dyn ClientTrait>`:
> 
> --> socketio/src/client/reconnect.rs:40:8 | 14 | pub trait ClientTrait { | ----------- this trait cannot be made into an object... ... 40 | fn emit<E, D>(&self, event: E, data: D) -> Result<()> | ^^^^ ...because method `emit` has generic type parameters ... 80 | fn emit_with_ack<F, E, D>( | ^^^^^^^^^^^^^ ...because method `emit_with_ack` has generic type parameters = help: consider moving `emit` to another trait = help: consider moving `emit_with_ack` to another trait
> 
> For more information about this error, try `rustc --explain E0038`. error: could not compile `rust_socketio` due to previous error

What is your definition of `ClientTrait`, and why do we necessarily need it? Anyways I agree with what @nshaaban-cPacket mentions and expose both versions of the client.

socketio/examples/callback.rs
SSebo: em... you are right

SSebo:
> > @1c3t3a Compiler does not complain until I create a trait object `Box<dyn ClientTrait>`:
> > --> socketio/src/client/reconnect.rs:40:8 | 14 | pub trait ClientTrait { | ----------- this trait cannot be made into an object... ... 40 | fn emit<E, D>(&self, event: E, data: D) -> Result<()> | ^^^^ ...because method `emit` has generic type parameters ... 80 | fn emit_with_ack<F, E, D>( | ^^^^^^^^^^^^^ ...because method `emit_with_ack` has generic type parameters = help: consider moving `emit` to another trait = help: consider moving `emit_with_ack` to another trait
> > For more information about this error, try `rustc --explain E0038`. error: could not compile `rust_socketio` due to previous error
> 
> What is your definition of `ClientTrait`, and why do we necessarily need it? Anyways I agree with what @nshaaban-cPacket mentions and expose both versions of the client.

If we expose both versions of client，we do not need this.

milandekruijf:
I've been testing this for a while now and it works great. I did notice that reconnects don't happen when losing internet connection, it just simply does nothing and stops emitting events.

SSebo:
> I've been testing this for a while now and it works great. I did notice that reconnects don't happen when losing internet connection, it just simply does nothing and stops emitting events.

@milandekruijf  @1c3t3a @nshaaban-cPacket  I found the js version of engine.io client will close connection if ping time out. https://github.com/socketio/engine.io-client/blob/dfee8ded722a8c4f9f773505d0c77b4561569863/lib/socket.ts#L645. Suggest @milandekruijf  to open a new issue, because the change will be in engine.io layer

TomieAi:
when this feature gonna be available? pls

nshaaban-cPacket:
What's left to do here:
-[ ] expose reconnecting client as Client in external API (returned from `connect` function on the ClientBuilder)
-[ ] expose Client as RawClient (or something along those lines) in public API
-[ ] expose second function on builder called `connect_raw` or something along those lines.
-[ ] bump MINOR version
-[ ] MERGE!

1c3t3a:
Sounds good! From my experience it was never bad advice to listen to @nshaaban-cPacket :) I would accept any PR that fixes the existing questions.

SSebo:
@1c3t3a @nshaaban-cPacket PTAL

socketio/examples/callback.rs
1c3t3a: Could you please remove that change? :)

1c3t3a (approved):
LGTM % comment :)

socketio/examples/callback.rs
SSebo: Of course.

socketio/src/client/client.rs
```diff
@@ -1,192 +1,52 @@
...
-/// soon as the ack'ed state is set to true. An `Ack` that didn't get ack'ed
-/// won't contain data.
-#[derive(Debug)]
-pub struct Ack {
-    pub id: i32,
-    timeout: Duration,
-    time_started: Instant,
-    callback: Callback<SocketCallback>,
-}
+#![allow(unused)]
```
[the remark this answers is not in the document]
SSebo: Of course.

socketio/src/client/client.rs
```diff
@@ -195,553 +55,245 @@ impl Client {
...
-        }
-        Ok(())
+        });
     }
 }
 
-pub struct Iter<'a> {
-    socket: &'a Client,
+pub(crate) struct Iter {
+    socket: Arc<RwLock<RawClient>>,
```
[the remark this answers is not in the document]
SSebo: RawClient is guarded by RwLock in Client, Can not just ref Client without hold the guard, when guard dropped, the ref is invlaid.

socketio/src/client/raw_client.rs
```diff
@@ -0,0 +1,748 @@
...
+                        println!("Ack data: {}", str);
+                    }
+                }
+            )
+            .is_ok());
+
+        Ok(())
+    }
+
+    // TODO: 0.3.X add secure socketio server
```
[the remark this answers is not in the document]
nshaaban-cPacket: Version is no longer relevant, not required for this PR, but would be nice to cleanup while we're here

socketio/src/client/client.rs
```diff
@@ -266,22 +338,19 @@ mod test {
...
         std::thread::sleep(Duration::from_millis(100));
-        let mut post_num = 0;
-        {
-            let packets = packets.read().unwrap();
-            post_num = packets.len();
-        }
+
+        let _packets = packets.read().unwrap();
+        let post_num = _packets.len();
+        drop(_packets);
```
nshaaban-cPacket, line 353: nit: _ prefix misleading
```suggestion
        let packets = packets.read().unwrap();
        let post_num = packets.len();
        drop(packets);
```

PR #236, opened 2022-09-06
feat: basic async socketio server

SSebo:
based on basic engine io server and async socketio client.

support join room, leave room,  emit to rooms

not support TLS.
not support multiple namespace.


tyilo:
@SSebo What's the reason for closing?

SSebo:
@tyilo I found some bugs. I will create a new PR when #180 is merged.

PR #237, opened 2022-09-29
docs: added code block closing

Ghamza-Jd:
Closing the code block

Screen Shot 2022-09-29 at 11 31 18 PM

### DIFF
--- a/socketio/src/client/builder.rs
+++ b/socketio/src/client/builder.rs
@@ -226,6 +226,7 @@ impl ClientBuilder {
     /// # Example
     /// ```rust
     /// use rust_socketio::{ClientBuilder, TransportType};
+    /// use serde_json::json;
     ///
     /// let socket = ClientBuilder::new("http://localhost:4200/")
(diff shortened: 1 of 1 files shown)


socketio/src/client/builder.rs
```diff
@@ -234,6 +234,12 @@ impl ClientBuilder {
     ///     .expect("connection failed");
     ///
     /// // use the socket
+    /// let json_payload = json!({"token": 123});
```
1c3t3a, line 237: If you use the `json!` macro here, please also add the import `use serde_json::json` as well.

1c3t3a (changes requested):
Good catch! Thanks for the PR :) I just have one comment.

socketio/src/client/builder.rs
Ghamza-Jd: Done, thanks!

1c3t3a (approved):
LGTM!

Issue #246, opened 2022-10-19
Emission with multiple parameters


david-sailplan:
The Client.emit and Client.emit_with_ack only seem to support a single data parameter

This causes issues for servers who are expecting multiple parameters for the data for these events

rust.rs
```
socket.emit_with_ack("foo", json!("{ack:true}", duration, callback) 
```

example_socketio_server.js
```
socket.on("foo", async (userId, body, callback)=>{})
```
no mater what is placed in data, it will place all data under the "userId" and no way to populate the body parameter.

Is there a workaround that could fix this?

1c3t3a:
Do you use a special [parser](https://socket.io/docs/v4/custom-parser/) for this on the server side?

david-sailplan:
No. We don't use any parser beyond the default behavior of the nodejs package server-side.

`const io = socketio(server, {
    path: '/socket',
    transports: ['websocket']
  })`

SSebo:
@david-sailplan https://github.com/1c3t3a/rust-socketio/blob/4ac27c7158a89375c503164857d86b1a1167116c/socketio/src/socket.rs#L105-L109 currently is limit to only two params, there is no workaround except add new function I think.

syshriki9:
@SSebo is there currently a better way to do this using "binary data/attachments"? 

I know your todo list is probably long, but a proper interface for encoding multiple message parameters would be a great feature :)

SSebo:
cc @1c3t3a @nshaaban-cPacket 

> @SSebo is there currently a better way to do this using "binary data/attachments"?
> 
> I know your todo list is probably long, but a proper interface for encoding multiple message parameters would be a great feature :)

cc @1c3t3a @nshaaban-cPacket

SSebo:
#249 I made a PR @david-sailplan You can have a try.

syshriki9:
Wow that was fast!, Thanks @SSebo.

As David mentioned, it seems to work.

syshriki9:
@SSebo @1c3t3a do you guys think this feature will make it into the master?

shenjackyuanjie:
also. that……
hmmm
well, now the `.on` can not accept mult payload, so ……
anther old issue

shenjackyuanjie:
This issue is mostly done, likely able to close

PR #249, opened 2022-10-24
feat: emit multipe parameters

SSebo:
js client can emit multiple parameters in one call. user socketio server `socket.on("foo", async (userId, body, callback)=>{})`  (#246) accept this kind of data which rust-socektio can not emit.

### DIFF
--- a/socketio/examples/callback.rs
+++ b/socketio/examples/callback.rs
@@ -11,8 +11,10 @@ fn main() {
     // socket to communicate with the server
     let handle_test = |payload: Payload, socket: RawClient| {
         match payload {
-            Payload::String(str) => println!("Received string: {}", str),
+            Payload::Json(json) => println!("Received string: {:?}", json),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
+            Payload::Multi(vec_data) => println!("Received vec: {:?}", vec_data),
+            _ => {}
         }
         socket
             .emit("test", json!({"got ack": true}))
--- a/socketio/examples/callback.rs
+++ b/socketio/examples/callback.rs
@@ -13,7 +13,6 @@ fn main() {
         match payload {
             Payload::String(str) => println!("Received string: {}", str),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
-            Payload::Number(num) => println!("Received number: {}", num),
         }
         socket
             .emit_multi("test", vec![json!({"got ack": true})])
--- a/socketio/examples/callback.rs
+++ b/socketio/examples/callback.rs
@@ -13,9 +13,10 @@ fn main() {
         match payload {
             Payload::String(str) => println!("Received string: {}", str),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
+            Payload::Number(num) => println!("Received number: {}", num),
         }
         socket
-            .emit("test", json!({"got ack": true}))
+            .emit_multi("test", vec![json!({"got ack": true})])
             .expect("Server unreachable")
     };
 
--- a/socketio/examples/callback.rs
+++ b/socketio/examples/callback.rs
@@ -15,7 +15,7 @@ fn main() {
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
         }
         socket
-            .emit_multi("test", vec![json!({"got ack": true})])
+            .emit("test", json!({"got ack": true}))
             .expect("Server unreachable")
     };
 
--- a/socketio/examples/readme.rs
+++ b/socketio/examples/readme.rs
@@ -8,8 +8,10 @@ fn main() {
     // socket to communicate with the server
     let callback = |payload: Payload, socket: RawClient| {
         match payload {
-            Payload::String(str) => println!("Received: {}", str),
+            Payload::Json(data) => println!("Received: {:?}", data),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
+            Payload::Multi(vec_data) => println!("Received vec: {:?}", vec_data),
+            _ => {}
         }
         socket
             .emit("test", json!({"got ack": true}))
--- a/socketio/examples/readme.rs
+++ b/socketio/examples/readme.rs
@@ -10,7 +10,6 @@ fn main() {
         match payload {
             Payload::String(str) => println!("Received: {}", str),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
-            Payload::Number(num) => println!("Received: {}", num),
         }
         socket
             .emit("test", json!({"got ack": true}))
--- a/socketio/examples/readme.rs
+++ b/socketio/examples/readme.rs
@@ -10,6 +10,7 @@ fn main() {
         match payload {
             Payload::String(str) => println!("Received: {}", str),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
+            Payload::Number(num) => println!("Received: {}", num),
         }
         socket
             .emit("test", json!({"got ack": true}))
--- a/socketio/src/client/client.rs
+++ b/socketio/src/client/client.rs
@@ -70,48 +70,11 @@ impl Client {
         client.emit(event, data)
     }
 
-    /// Sends a message to the server using the underlying `engine.io` protocol.
-    /// This message takes an event, which could either be one of the common
-    /// events like "message" or "error" or a custom event like "foo". But be
-    /// careful, the data string needs to be valid JSON. It's recommended to use
-    /// a library like `serde_json` to serialize the data properly.
-    ///
-    /// # Example
-    /// ```
-    /// use rust_socketio::{ClientBuilder, RawClient, Payload};
-    /// use serde_json::json;
-    ///
-    /// let mut socket = ClientBuilder::new("http://localhost:4200/")
-    ///     .on("test", |payload: Payload, socket: RawClient| {
-    ///         println!("Received: {:#?}", payload);
-    ///         socket.emit("test", json!({"hello": true})).expect("Server unreachable");
-    ///      })
-    ///     .connect()
-    ///     .expect("connection failed");
-    ///
-    /// let payload = vec![json!({"token": 123})];
-    ///
-    /// let result = socket.emit_multi("foo", payload);
-    ///
-    /// assert!(result.is_ok());
-    /// ```
-    pub fn emit_multi<E, D>(&self, event: E, data: Vec<D>) -> Result<()>
-    where
-        E: Into<Event>,
-        D: Into<Payload>,
-    {
-        let client = self.client.read()?;
-        // TODO(#230): like js client, buffer emit, resend after reconnect
-        client.emit_multi(event, data)
-    }
-
     /// Sends a message to the server but `alloc`s an `ack` to check whether the
     /// server responded in a given time span. This message takes an event, which
     /// could either be one of the common events like "message" or "error" or a
-    /// custom event like "foo", as well as a data parameter. But be careful,
-    /// in case you send a [`Payload::String`], the string needs to be valid JSON.
-    /// It's even recommended to use a library like serde_json to serialize the data properly.
-    /// It also requires a timeout `Duration` in which the client needs to answer.
+    /// custom event like "foo", as well as a data parameter.
+    /// It requires a timeout `Duration` in which the client needs to answer.
     /// If the ack is acked in the correct time span, the specified callback is
     /// called. The callback consumes a [`Payload`] which represents the data send
     /// by the server.
@@ -130,8 +93,10 @@ impl Client {
     ///
     /// let ack_callback = |message: Payload, socket: RawClient| {
     ///     match message {
-    ///         Payload::String(str) => println!("{}", str),
+    ///         Payload::Json(data) => println!("{:?}", data),
     ///         Payload::Binary(bytes) => println!("Received bytes: {:#?}", bytes),
+    ///         Payload::Multi(vec) => println!("Received multi: {:?}", vec),
+    ///         _ => {}
     ///    }
     /// };
     ///
@@ -155,58 +120,6 @@ impl Client {
         let client = self.client.read()?;
         // TODO(#230): like js client, buffer emit, resend after reconnect
         client.emit_with_ack(event, data, timeout, callback)
-    }
-
-    /// Sends a message to the server but `alloc`s an `ack` to check whether the
-    /// server responded in a given time span. This message takes an event, which
-    /// could either be one of the common events like "message" or "error" or a
-    /// custom event like "foo", as well as a data parameter. But be careful,
-    /// in case you send a [`Payload::String`], the string needs to be valid JSON.
-    /// It's even recommended to use a library like serde_json to serialize the data properly.
-    /// It also requires a timeout `Duration` in which the client needs to answer.
-    /// If the ack is acked in the correct time span, the specified callback is
-    /// called. The callback consumes a [`Payload`] which represents the data send
-    /// by the server.
-    ///
-    /// # Example
-    /// ```
-    /// use rust_socketio::{ClientBuilder, Payload, RawClient};
-    /// use serde_json::json;
-    /// use std::time::Duration;
-    /// use std::thread::sleep;
-    ///
-    /// let mut socket = ClientBuilder::new("http://localhost:4200/")
-    ///     .on("foo", |payload: Payload, _| println!("Received: {:#?}", payload))
-    ///     .connect()
-    ///     .expect("connection failed");
-    ///
-    /// let ack_callback = |message: Payload, socket: RawClient| {
-    ///     match message {
-    ///         Payload::String(str) => println!("{}", str),
-    ///         Payload::Binary(bytes) => println!("Received bytes: {:#?}", bytes),
-    ///    }
-    /// };
-    ///
-    /// let payload = vec![json!({"token": 123})];
-    /// socket.emit_multi_with_ack("foo", payload, Duration::from_secs(2), ack_callback).unwrap();
-    ///
-    /// sleep(Duration::from_secs(2));
-    /// ```
-    pub fn emit_multi_with_ack<F, E, D>(
-        &self,
-        event: E,
-        data: Vec<D>,
-        timeout: Duration,
-        callback: F,
-    ) -> Result<()>
-    where
-        F: for<'a> FnMut(Payload, RawClient) + 'static + Sync + Send,
-        E: Into<Event>,
-        D: Into<Payload>,
-    {
-        let client = self.client.read()?;
-        // TODO(#230): like js client, buffer emit, resend after reconnect
-        client.emit_multi_with_ack(event, data, timeout, callback)
     }
 
     /// Disconnects this client from the server by sending a `socket.io` closing
--- a/socketio/src/client/client.rs
+++ b/socketio/src/client/client.rs
@@ -132,7 +132,6 @@ impl Client {
     ///     match message {
     ///         Payload::String(str) => println!("{}", str),
     ///         Payload::Binary(bytes) => println!("Received bytes: {:#?}", bytes),
-    ///         Payload::Number(num) => println!("Received number: {}", num),
     ///    }
     /// };
     ///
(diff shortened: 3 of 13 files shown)


david-sailplan:
Seems to be working as expected for the most part but I seem to be running into issues when the Value is a string. Namely it seems to not take that value as a singular argument, but instead as an array of characters and sends each character as a different argument.

Of course, I might be doing something wrong here, so please correct me if I am

For example lets say I have 
```
socket.on("test", async(arg1, arg2, arg3, callback)=>{
   console.log([arg1, arg2, arg3])
});
```

The following will work as expected 
```
let args = vec![json!({"foo": "foo"}), json!("{bar": "bar"}), json!({"foobar": "foobar"})]
client.emit_multi_with_ack("test", args, duration, callback);
```
Result
```
[{"foo": "foo"}, {"bar": "bar"}, {"foobar": "foobar"}]
```


But if you try to send a string it seems to split the arguments into the characters
```
let args = vec![json!("foo"), json!("bar"), json!("foobar")]
client.emit_multi_with_ack("test", args, duration, callback);
```

Results in the following
```
['f', 'o', 'o']
```

socketio/examples/callback.rs
```diff
@@ -13,9 +13,10 @@ fn main() {
         match payload {
             Payload::String(str) => println!("Received string: {}", str),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
+            Payload::Number(num) => println!("Received number: {}", num),
```
nshaaban-cPacket, line 16: What value does this type add? Isn't it already covered by binary?

socketio/src/client/raw_client.rs
```diff
@@ -200,9 +236,86 @@ impl RawClient {
...
+    ///    }
+    /// };
+    ///
+    /// let payload = json!({"token": 123});
+    /// socket.emit_with_ack("foo", payload, Duration::from_secs(2), ack_callback).unwrap();
+    ///
+    /// sleep(Duration::from_secs(2));
+    /// ```
+    #[inline]
+    pub fn emit_multi_with_ack<F, E, D>(
```
nshaaban-cPacket, line 298: What value does this add? Could we add and `impl From Vec<...> for Payload `, so the existing function works?

socketio/src/socket.rs
```diff
@@ -70,54 +70,77 @@ impl Socket {
...
-
-                Ok(Packet::new(
-                    PacketId::Event,
-                    nsp.to_owned(),
-                    Some(payload),
-                    id,
-                    0,
-                    None,
-                ))
+        let mut attachments = vec![];
```
nshaaban-cPacket, line 95: nit/suggestion: code is hard to follow

socketio/src/packet.rs
```diff
@@ -118,24 +118,8 @@ impl From<&Packet> for Bytes {
...
-                format!(
-                    "[{{\"_placeholder\":true,\"num\":{}}}]",
-                    packet.attachment_count - 1,
-                )
-            };
 
-            // build the buffers
-            buffer.put(placeholder.as_ref());
-        } else if let Some(data) = packet.data.as_ref() {
+        if let Some(data) = packet.data.as_ref() {
```
nshaaban-cPacket, line 122: these two are not equivalent, what changed?

socketio/examples/readme.rs
```diff
@@ -10,6 +10,7 @@ fn main() {
         match payload {
             Payload::String(str) => println!("Received: {}", str),
             Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
+            Payload::Number(num) => println!("Received: {}", num),
```
nshaaban-cPacket, line 13: nit: what value does this add?

nshaaban-cPacket (commented):
Thanks for the PR! Don't quite understand what the problem was, nor the solution, a brief explanation (preferably in comments) as to what happened may be helpful

david-sailplan:
I had left an earlier comment but deleted it. Turns out it was an issue on the server side

david-sailplan:
This does seem to solve the problem I was experiencing [here](https://github.com/1c3t3a/rust-socketio/issues/246#issuecomment-1286424751). Good job

socketio/examples/callback.rs
SSebo: currently rust_socketio can not emit `number` value, it will encode as string

socketio/src/client/raw_client.rs
SSebo: This function wants to emits `Vec<Payload>` as what javascript version of socketio client did.

socketio/src/packet.rs
SSebo: currently only support emit one binary payload, if we want to emit multiple binary payload and keep the same order user emit, placeholders are hard to replace in `packet.rs`. so this logic is moved to `socket.rs`

socketio/examples/readme.rs
SSebo: currently can not emit simple number payload

1c3t3a:
First of all thanks for the quick solution @SSebo! But I wanted to ask whether this can be solved by reverting the behaviour of #220. This was the initial reason why I only accepted valid json and nothing else. I would prefer this approach over a completely new API call!

1c3t3a:
This is not entirely true, we don't need number since this is part of Json, eg: `{ "my_number": 1}`. Can you confirm that a revert of #220 results in being able to emit to a socketio server with `socket.on("foo", async (userId, body, callback)=>{})`where body is not empty? :)

SSebo:
> 

I agree, not include number, I removed Payload::Number part.

SSebo:
> This is not entirely true, we don't need number since this is part of Json, eg: `{ "my_number": 1}`. Can you confirm that a revert of #220 results in being able to emit to a socketio server with `socket.on("foo", async (userId, body, callback)=>{})`where body is not empty? :)

@david-sailplan can you confirm that use a single json payload is able to emit to a socketio server with `socket.on("foo", async (userId, body, callback)=>{})`where body is not empty? :)

david-sailplan:
@SSebo I'm not entirely sure what you're asking here. If you're asking if something to the effect of 
```
let payload = json!(("user_id_1", json!({'data": 1})))
socket.emit_with_ack("foo",payload, duration, callback) 
```
works with server code
```
socket.on("foo", async (userId, body, callback)=>{
    console.log({userId, body, callback});
    callback(1);
})
```

I can confirm that it does not. Log would be 
```
{
    userId: ['user_id)1', {'data': 1}]
    body: [Function (anonymous)],
    callback: undefined
}
```

david-sailplan:
If what you're asking is if 
```
let payload = vec![json!(1), json({"data": 2}))]
socket.emit_multi_with_ack("foo", payload, duration, callback)
```
works with the above server code the answer is yes. It behaves as we expected even after removing the Payload::Number

socketio/src/socket.rs
```diff
@@ -70,52 +70,85 @@ impl Socket {
...
+        data += "]";
+
+        (data, attachments)
+    }
+
+    fn encode_payloads(data: &mut String, payloads: Vec<Payload>, attachments: &mut Vec<Bytes>) {
+        for (index, payload) in payloads.iter().enumerate() {
+            match payload {
+                Payload::Binary(bin_data) => {
+                    *data += "{\"_placeholder\":true,\"num\":";
```
nshaaban-cPacket, line 136: nit: should this number of attachments be done at the top level?

socketio/src/socket.rs
```diff
@@ -70,52 +70,85 @@ impl Socket {
...
+        let mut data = "[".to_owned();
+
+        if let Some(event) = event {
+            data += &format!("\"{}\"", String::from(event));
+            if !payloads.is_empty() {
+                data += ","
+            }
+        }
+
+        Self::encode_payloads(&mut data, payloads, &mut attachments);
```
nshaaban-cPacket, line 125: nit: seems to be re-implementing a json serialization. Would storing in a temporary struct then use serde_json's to_string be simpler?

socketio/src/client/client.rs
```diff
@@ -122,6 +157,58 @@ impl Client {
...
+    ///         Payload::Binary(bytes) => println!("Received bytes: {:#?}", bytes),
+    ///    }
+    /// };
+    ///
+    /// let payload = vec![json!({"token": 123})];
+    /// socket.emit_multi_with_ack("foo", payload, Duration::from_secs(2), ack_callback).unwrap();
+    ///
+    /// sleep(Duration::from_secs(2));
+    /// ```
+    pub fn emit_multi_with_ack<F, E, D>(
```
nshaaban-cPacket, line 195: nit: we may be able to avoid the need for a new function by modifying the payload type, perhaps adding a Multi variant to payload?

nshaaban-cPacket (commented):
Just dropping by, as always @1c3t3a 's in charge

socketio/src/payload.rs
```diff
@@ -1,31 +1,45 @@
...
-/// used for both representing data that's send and data that's received.
+/// The enum is used for both representing data that's send and
+/// data that's received.
 #[derive(Debug, PartialEq, Eq, Clone)]
+#[non_exhaustive]
 pub enum Payload {
     Binary(Bytes),
-    String(String),
+    Json(Value),
+    Multi(Vec<RawPayload>),
```
nshaaban-cPacket, line 12: Not sure if this is better, or if this type even should be recursive, another option would be:
```suggestion
    Multi(Vec<Box<Payload>>),
```

socketio/src/socket.rs
```diff
@@ -70,54 +87,75 @@ impl Socket {
...
     /// Returns a packet for a payload, could be used for bot binary and non binary
     /// events and acks. Convenance method.
     #[inline]
-    pub(crate) fn build_packet_for_payload<'a>(
-        &'a self,
+    pub(crate) fn build_packet_for_payload(
         payload: Payload,
-        event: Event,
-        nsp: &'a str,
+        event: Option<Event>,
```
nshaaban-cPacket, line 101: nit: when is event none? Only usage I can see is in the tests

socketio/src/socket.rs
```diff
@@ -16,6 +18,21 @@ pub(crate) struct Socket {
     connected: Arc<AtomicBool>,
 }
 
+#[derive(Serialize)]
+struct BinaryPlaceHolder {
```
nshaaban-cPacket, line 22: :heart:

socketio/src/packet.rs
```diff
@@ -234,27 +220,12 @@ impl TryFrom<&Bytes> for Packet {
...
-                if str.is_empty() {
-                    packet.data = None
-                } else {
-                    packet.data = Some(str)
-                }
-            }
-            _ => packet.data = Some(json_data.to_string()),
+        packet.data = match json_data {
+            Value::Array(vec) if vec.is_empty() => None,
+            // Value::Array(vec) if vec.len() == 1 => vec.get(0).map(|v| v.to_owned()),
```
nshaaban-cPacket, line 227:
```suggestion
```

socketio/src/packet.rs
```diff
@@ -234,27 +220,12 @@ impl TryFrom<&Bytes> for Packet {
...
+            _ => Some(json_data),
```
nshaaban-cPacket, line 228: nit: would this still have the `_placeholder` in it?

socketio/src/packet.rs
```diff
@@ -21,7 +21,7 @@ pub enum PacketId {
 pub struct Packet {
     pub packet_type: PacketId,
     pub nsp: String,
-    pub data: Option<String>,
+    pub data: Option<Value>,
```
nshaaban-cPacket, line 24: blocker: This is a public type, changing this would be a breaking change. While I agree data: Option<Value> makes the internal implementation more elegant, not worth introducing more breaking changes. Changing the payload type on a PATCH IMO is already pushing the boundary of what should be done on a PATCH even in 0.X.X
```suggestion
    // FIXME: on next breaking change consider changing this to a Value
    pub data: Option<String>,
```

socketio/src/payload.rs
```diff
@@ -1,31 +1,45 @@
...
-/// `Payload::String` which holds a [`std::string::String`]. The enum is
-/// used for both representing data that's send and data that's received.
+/// The enum is used for both representing data that's send and
+/// data that's received.
 #[derive(Debug, PartialEq, Eq, Clone)]
+#[non_exhaustive]
 pub enum Payload {
     Binary(Bytes),
-    String(String),
+    Json(Value),
```
nshaaban-cPacket, line 11: nit: consider adding a way to have some sort of way to convert a struct that impls Deserialize into Payload.

socketio/src/client/raw_client.rs
```diff
@@ -306,51 +319,95 @@ impl RawClient {
...
+            },
+            _ => Event::Message,
         };
 
-        if let Some(attachments) = &packet.attachments {
-            if let Some(binary_payload) = attachments.get(0) {
-                self.callback(&event, Payload::Binary(binary_payload.to_owned()))?;
+        let payload = match &packet.data {
+            Some(Value::Array(vec)) if vec.len() == 1 => {
+                let value = vec.get(0).unwrap();
```
nshaaban-cPacket, line 333:
```suggestion
                // SAFETY: length checked above
                let value = vec.get(0).unwrap();
```

socketio/src/client/raw_client.rs
```diff
@@ -306,51 +319,95 @@ impl RawClient {
...
+                _ => Event::Message,
+            },
+            _ => Event::Message,
         };
 
-        if let Some(attachments) = &packet.attachments {
-            if let Some(binary_payload) = attachments.get(0) {
-                self.callback(&event, Payload::Binary(binary_payload.to_owned()))?;
+        let payload = match &packet.data {
+            Some(Value::Array(vec)) if vec.len() == 1 => {
```
nshaaban-cPacket, line 332: nit: clippy lint https://rust-lang.github.io/rust-clippy/master/#len_zero
```suggestion
            Some(Value::Array(vec)) if ! vec.is_empty() => {
```

socketio/src/client/raw_client.rs
```diff
@@ -306,51 +319,95 @@ impl RawClient {
...
+                        serde_json::Value::String(ev) => ev,
+                        _ => "message",
+                    })
+                    .unwrap_or("message")
+                    .into()
+            } else {
+                Event::Message
+            };
+
+            let payload = if contents.len() > 1 {
```
nshaaban-cPacket, line 398: nit: cilppy lint
```suggestion
            let payload = if ! contents.is_empty() {
```

socketio/src/client/raw_client.rs
```diff
@@ -306,51 +319,95 @@ impl RawClient {
...
+                Payload::Multi(contents[1..].iter().map(|v| v.to_owned().into()).collect())
```
nshaaban-cPacket, line 399: nit: alternative. No difference at all, just if you feel like using iterator's functions
```suggestion
                Payload::Multi(contents.iter().skip(1).cloned().map(RawPayload::from).collect())
```

socketio/src/client/raw_client.rs
```diff
@@ -306,51 +319,95 @@ impl RawClient {
...
+        // the string must be a valid json array with the event at index 0 and the
+        // payload at index 1. if no event is specified, the message callback is used
+        if let Some(serde_json::Value::Array(contents)) = &packet.data {
+            let event: Event = if contents.len() > 1 {
+                contents
+                    .get(0)
+                    .map(|value| match value {
+                        serde_json::Value::String(ev) => ev,
+                        _ => "message",
+                    })
```
nshaaban-cPacket, line 391:
```suggestion
                    .and_then(|value| match value {
                        serde_json::Value::String(ev) => Some(ev),
                        _ => None
                    })
```

nshaaban-cPacket (commented):
Thank you for being receptive to comments so far, feel free to ignore the "nit"s you don't want to do, just saying things I noticed.

Changes to Packet's public interface require @1c3t3a 's approval. We MIGHT be able to still squeeze by with changes to Packet, as it's not likely many use it. Your call.

socketio/src/payload.rs
SSebo: I tried `Multi(Vec<Payload>)` first time, recursive actually can be convert to valid json, but not what we expected. To not let user do that, `RawPayload` is provided. also let code simple.

socketio/src/socket.rs
SSebo: I found  `client.emit(Buffer.from([4, 5, 6]));` in socket-io.js, I thought event is optional that time. But I lookup the socketio dos, event is not optional. Does this mean the event can be a valid payload?

socketio/src/packet.rs
SSebo: Yes, _placeholder will be replaced to real `Payload::Binary` in socket.rs

socketio/src/packet.rs
SSebo: Since the `Payload::Multi` is a breaking change, so I change data to `Value`.

socketio/src/client/raw_client.rs
SSebo: `len() > 0`  is equal to `! is_empty`. Here is `len() > 1`, is just one item, no need to use `Payload::Multi`

socketio/src/socket.rs
SSebo: right now in `rust-socketio` if only receive 1 item in packet.data, it will be treat as payload, not event.

https://github.com/1c3t3a/rust-socketio/blob/4ac27c7158a89375c503164857d86b1a1167116c/socketio/src/client/raw_client.rs#L325-L354

socketio/src/client/raw_client.rs
SSebo: here is len() == 1

socketio/src/socket.rs
1c3t3a: Where did you find that? Specification only mentions, that when no event is specified, "Message" should be used. So having an `Option<Event>` would not make sense in that case.

socketio/src/payload.rs
1c3t3a: I think this makes it a bit obvious that restructuring payload is probably a bad idea. What you do here is basically rewrite a limited Json `Value` enum, with support for binary data. I would rather leverage json, and solve this problem differently. I am going to take a look at this tomorrow, but I am not really in favor of such changes, as they present obvious problems.

socketio/src/packet.rs
nshaaban-cPacket: packet.rs should handle all de-serialization of the packet. This will become more important if/when we want to support multiple versions.

socketio/src/client/raw_client.rs
nshaaban-cPacket: Yes hence why it's `! vec.is_empty()`. is_empty() is as faster operation than len()

socketio/src/socket.rs
```diff
@@ -158,6 +158,47 @@ impl Socket {
         attachments.push(bin_data);
     }
 
+    pub(crate) fn decode_binary_payload(
```
nshaaban-cPacket, line 161: nit: why did this move? It's only used in raw_client

SSebo: I found encode is in socket.rs, so I move decode in socket.rs too.

socketio/src/socket.rs
SSebo: @1c3t3a I found it here https://github.com/1c3t3a/rust-socketio/blob/4ac27c7158a89375c503164857d86b1a1167116c/ci/socket-io.js#L41

socketio/src/client/raw_client.rs
SSebo: I got that, if not empty, get(0) will not panic.

socketio/src/packet.rs
SSebo: `BinaryEvent` can be mixed with normal data and binary data like `client.emit('test', Buffer.from([1, 2, 3]), "4", Buffer.from([5, 6]));`, so we can not just remove all `_placeholder` as we do not know where the binary data is in the sequence of received items. `socket.rs` not parse the packet, just replace `_placehoulder` to `Payload::Binary`

nshaaban-cPacket:
Why was this closed @SSebo ?

1c3t3a:
Good question!

PR #250, opened 2022-11-01
Fix missing space in ci script

tyilo:


1c3t3a (approved):
Thanks for catching this! :)

Issue #251, opened 2022-11-01
Set url parameters for reconnecting


tyilo:
I need to connect a socket.io that on connection will check if a specific query parameter contains a valid JWT token (which is not expired).

During the lifetime of the client using this library, the JWT token will expire. I can retrieve a fresh token which is not expired, but I need to update the url that rust-socketio uses for reconnection, otherwise reconnection will fail.

I see two different approaches for implementing this:
1. Add a public `set_reconnect_url` function to `Client`.
2. Add a `on_reconnect` function to `ClientBuilder` with the signature `fn(ClientBuilder) -> ClientBuilder`. Before the client tries to reconnect it will use the reconnect handler (if specified) to get a modified `ClientBuilder`. This handler can then change the url, opening headers, auth, etc...

I have implemented option 1 in https://github.com/1c3t3a/rust-socketio/pull/252

1c3t3a:
Ough.. I think this is really odd behavior that doesn't really sound like proper API design. Have you tried using something like HTTP 301 status codes to do redirection to different endpoint? I am not sure if I like the idea of redesigning the public interface for this "edge-casey" use case.

tyilo:
> Ough.. I think this is really odd behavior that doesn't really sound like proper API design. Have you tried using something like HTTP 301 status codes to do redirection to different endpoint? I am not sure if I like the idea of redesigning the public interface for this "edge-casey" use case.

That doesn't seem to work with authentication.

Also the javascript socket.io library supports this.

1c3t3a:
Can you point me to the JS implementation of this?

tyilo:
The official socket.io documentation has an example of updating the query parameters on reconnect[^1]:
```js
socket.io.on("reconnect_attempt", () => {
  socket.io.opts.query.x++;
});
```

I actually also only need the ability to change the query parameters, not the full url.

[^1]: https://socket.io/docs/v4/client-options/#query

rageshkrishna:
This seems to be a valid use case. I'm running into the same auth problem myself now. When a reconnect is required, I should be able to pass a new auth value. It seems like a callback for reconnect would solve this and a few other problems, like the app needing to know when a reconnect is about to happen.

tyilo:
@rageshkrishna I think you should open a new issue, as this issue has been fixed.

PR #252, opened 2022-11-01
Add set_reconnect_url to Client

tyilo:


1c3t3a:
Sorry for the late review and thanks for the patch! I don't really understand this patch though, why do we need this instead of setting a different address when we create the builder? This doesn't give any benefit in my eyes.

tyilo:
> Sorry for the late review and thanks for the patch! I don't really understand this patch though, why do we need this instead of setting a different address when we create the builder? This doesn't give any benefit in my eyes.

I forgot to link it, but this fixes #251 where I have described why I need this feature.

tyilo:
Ping @1c3t3a

Cargo.lock
1c3t3a, line 1: You shouldn't have changes in this file if you didn't add any dependency, please check for that! :)

socketio/src/client/client.rs
```diff
@@ -35,6 +35,10 @@ impl Client {
         Ok(s)
     }
 
+    pub fn set_reconnect_url<T: Into<String>>(&self, address: T) {
```
1c3t3a, line 38: Can you please add a doc comment how this method works and why it exists?

ci/socket-io-restart-url-auth.js
```diff
@@ -0,0 +1,40 @@
...
+const port = 4206;
+const timeout = 200;
+
+function isTimestampValid(timestampStr) {
+  if (timestampStr === undefined) return false;
+  const timestamp = parseInt(timestampStr);
+  if (isNaN(timestamp)) return false;
+
+  const diff = Date.now() - timestamp;
+  return Math.abs(diff) <= 1000;
```
1c3t3a, line 15: Please add a comment on why this number is 1000 or store it in a `const` with a descriptive name.

ci/socket-io-restart-url-auth.js
```diff
@@ -0,0 +1,40 @@
+const { Socket } = require("socket.io");
+
+let createServer = require("http").createServer;
+let server = createServer();
+const io = require("socket.io")(server);
+const port = 4206;
+const timeout = 200;
+
+function isTimestampValid(timestampStr) {
```
1c3t3a, line 9: nit: isValidTimestamp

socketio/src/client/client.rs
```diff
@@ -308,6 +322,84 @@ mod test {
...
+            let timestamp = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_millis();
+            let mut url = crate::test::socket_io_restart_url_auth_server();
+            url.set_query(Some(&format!("timestamp={timestamp}")));
+            url
+        }
+
+        let connect_num = Arc::new(AtomicUsize::new(0));
```
1c3t3a, line 337: Does this really need to wrapped into an `Arc`? Can't we send the atomics across the thread boundary?

socketio/src/client/client.rs
```diff
@@ -308,6 +322,84 @@ mod test {
...
+        let close_num_clone = Arc::clone(&close_num);
+        let message_num_clone = Arc::clone(&message_num);
+
+        let socket = ClientBuilder::new(get_url())
+            .reconnect(true)
+            .max_reconnect_attempts(100)
+            .reconnect_delay(100, 100)
+            .on(Event::Connect, move |_, socket| {
+                connect_num_clone.fetch_add(1, Ordering::SeqCst);
+                let r = socket.emit_with_ack(
```
1c3t3a, line 351: nit: Please pick a more descriptive name than "r"

socketio/src/client/client.rs
```diff
@@ -308,6 +322,84 @@ mod test {
...
+                    json!(""),
+                    Duration::from_millis(100),
+                    |_, _| {},
+                );
+                assert!(r.is_ok(), "should emit message success");
+            })
+            .on(Event::Close, move |_, _| {
+                close_num_clone.fetch_add(1, Ordering::SeqCst);
+            })
+            .on("message", move |_, _socket| {
```
1c3t3a, line 362: the arguments to the closure can be `move |_, _|`. No need to name the socket.

socketio/src/client/client.rs
```diff
@@ -308,6 +322,84 @@ mod test {
...
+        assert_eq!(load(&connect_num), 1, "should connect once");
+        assert_eq!(load(&message_num), 1, "should receive one");
+        assert_eq!(load(&close_num), 0, "should not close");
+
+        // waiting for timestamp in url to expire
+        std::thread::sleep(std::time::Duration::from_secs(1));
+
+        socket.set_reconnect_url(get_url());
+
+        let r = socket.emit("restart_server", json!(""));
```
1c3t3a, line 384: nit: Please don't name something "r"

1c3t3a (changes requested):

1c3t3a:
Sorry for the delay!! I have some comments :) Can you please also rebase and resolve the merge conflicts?

socketio/src/client/client.rs
tyilo: Yes, we need to access it from multiple threads so it must be behind an `Arc`.

tyilo:
Rebased and updated the code.

socketio/src/client/client.rs
```diff
@@ -322,6 +338,84 @@ mod test {
...
+        let connect_num_clone = Arc::clone(&connect_num);
+        let close_num_clone = Arc::clone(&close_num);
+        let message_num_clone = Arc::clone(&message_num);
+
+        let socket = ClientBuilder::new(get_url())
+            .reconnect(true)
+            .max_reconnect_attempts(100)
+            .reconnect_delay(100, 100)
+            .on(Event::Connect, move |_, socket| {
+                connect_num_clone.fetch_add(1, Ordering::SeqCst);
```
1c3t3a, line 366: This can be `Acquire` and `Release´ with the load operation.

socketio/src/client/client.rs
```diff
@@ -210,7 +222,7 @@ impl Client {
                     Ok(Packet {
                         packet_type: PacketId::Disconnect,
                         ..
-                    }) => self_clone.builder.reconnect_on_disconnect,
+                    }) => self_clone.builder.lock().unwrap().reconnect_on_disconnect,
```
1c3t3a, line 225: Why unwrap? You can use `?` here.

socketio/src/client/client.rs
```diff
@@ -158,9 +164,14 @@ impl Client {
 
     fn reconnect(&mut self) {
         let mut reconnect_attempts = 0;
-        if self.builder.reconnect {
+        let (reconnect, max_reconnect_attempts) = {
+            let builder = self.builder.lock().unwrap();
```
1c3t3a, line 168: Don't unwrap, instead return a `Result<()>` and use `?`

socketio/src/client/client.rs
```diff
@@ -39,6 +39,12 @@ impl Client {
         Ok(s)
     }
 
+    /// Updates the URL the client will connect to when reconnecting.
+    /// This is especially useful for updating query parameters.
+    pub fn set_reconnect_url<T: Into<String>>(&self, address: T) {
+        self.builder.lock().unwrap().address = address.into();
```
1c3t3a, line 45: No unwrap please

socketio/src/client/client.rs
```diff
@@ -322,6 +338,84 @@ mod test {
...
+            let timestamp = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_millis();
+            let mut url = crate::test::socket_io_restart_url_auth_server();
+            url.set_query(Some(&format!("timestamp={timestamp}")));
+            url
+        }
+
+        let connect_num = Arc::new(AtomicUsize::new(0));
```
1c3t3a, line 353: Please make them static, you should be able to drop the Arc then.

socketio/src/client/client.rs
```diff
@@ -322,6 +338,84 @@ mod test {
...
+                // stream of packets, even when reconnecting
+                message_num_clone.fetch_add(1, Ordering::SeqCst);
+            })
+            .connect();
+
+        assert!(socket.is_ok(), "should connect success");
+        let socket = socket.unwrap();
+
+        // waiting for server to emit message
+        std::thread::sleep(std::time::Duration::from_millis(500));
```
1c3t3a, line 389: Please rather use a channel to signal you're done, this can lead to timing issues in CI.

socketio/src/client/client.rs
```diff
@@ -322,6 +338,84 @@ mod test {
...
+
+        assert_eq!(load(&connect_num), 1, "should connect once");
+        assert_eq!(load(&message_num), 1, "should receive one");
+        assert_eq!(load(&close_num), 0, "should not close");
+
+        // waiting for timestamp in url to expire
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
1c3t3a, line 396: Same as above, please use a channel or another signal.

socketio/src/client/client.rs
tyilo: Does it really matter for the test? The test is based on the existing one which also uses `SeqCst`.

socketio/src/client/client.rs
tyilo: Again, is this really important when this is just for a test? This uses `Arc`'s like the existing test in the file.

socketio/src/client/client.rs
tyilo: This is the same as the existing test. We have to wait a bit for the server to emit a message.

The (nodejs) server can't send a signal to the Rust code using a channel.

socketio/src/client/client.rs
tyilo: Same reply as above.

socketio/src/client/client.rs
1c3t3a: I think yes. If you treat your test code badly you will run into trouble sooner or later. Also "someone else made the same mistake" is a bad argument for code IMO.

socketio/src/client/client.rs
1c3t3a: I think yes. If you treat your test code badly you will run into trouble sooner or later. Also "someone else made the same mistake" is a bad argument for code IMO.

socketio/src/client/client.rs
1c3t3a: Ok, then should be fine for now.

socketio/src/client/client.rs
tyilo: Ok, it can't really be `Release` for the load operation:

```
error: atomic loads cannot have `Release` or `AcqRel` ordering
   --> socketio/src/client/client.rs:480:18
    |
480 |         num.load(Ordering::Release)
    |                  ^^^^^^^^^^^^^^^^^
    |
    = help: consider using ordering modes `Acquire`, `SeqCst` or `Relaxed`
    = note: `#[deny(invalid_atomic_ordering)]` on by default
```

So what should we use?

1c3t3a: Sorry, load should be `Acquire`, store `Release`.

tyilo: I think it might make the tests simpler to just use a `Mutex<usize>` instead of a `AtomicUsize`.

1c3t3a: Why? I would always prefer an atomic over a Mutex.

socketio/src/client/client.rs
tyilo: Fair enough.

I've changed the code and agree that looks better :)

socketio/src/client/client.rs
tyilo: Ok, I've changed the orderings.

1c3t3a (approved):
LGTM :) Thanks you!

tyilo:
Thanks for merging :)

PR #264, opened 2022-12-15
implement support for `?Sync` callbacks

### LINKED ISSUE #263: `ClientBuilder::on(_any)` should be able to take `!Sync` functions too
There's no reason to keep callbacks under `std::sync::RwLock` if it's `FnMut`, so exclusive access is needed anyways. `std::sync::Mutex` should be used instead as it implements `Sync` regardles if the wrapped value implements `Sync` or not.

ZetaNumbers:
Resolves #263

### DIFF
--- a/socketio/src/client/builder.rs
+++ b/socketio/src/client/builder.rs
@@ -10,7 +10,7 @@ use url::Url;
 use crate::client::callback::{SocketAnyCallback, SocketCallback};
 use crate::error::Result;
 use std::collections::HashMap;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex};
 
 use crate::socket::Socket as InnerSocket;
 
@@ -34,8 +34,8 @@ pub enum TransportType {
 #[derive(Clone)]
 pub struct ClientBuilder {
     address: String,
-    on: Arc<RwLock<HashMap<Event, Callback<SocketCallback>>>>,
-    on_any: Arc<RwLock<Option<Callback<SocketAnyCallback>>>>,
+    on: Arc<Mutex<HashMap<Event, Callback<SocketCallback>>>>,
+    on_any: Arc<Mutex<Option<Callback<SocketAnyCallback>>>>,
     namespace: String,
     tls_config: Option<TlsConnector>,
     opening_headers: Option<HeaderMap>,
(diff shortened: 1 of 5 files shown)


socketio/src/client/builder.rs
```diff
@@ -177,11 +177,11 @@ impl ClientBuilder {
     #[allow(unused_mut)]
```
1c3t3a, line 177: We can remove this now I assume?

socketio/src/client/builder.rs
```diff
@@ -149,11 +149,11 @@ impl ClientBuilder {
...
```
1c3t3a, line 149: Same as below, mutex lock should take a `mut self`.

1c3t3a (approved):
LGTM % comments. Thanks for the patch :)

socketio/src/client/builder.rs
ZetaNumbers: not sure what you meant but mut is indeed absolutelly useless, so i removed it

ZetaNumbers: Perhaps arc and mutex (previously rwlock) are created pretty early. Maybe we could create arc and mutex inside on connect method instead, but builder whould work with pure `Option<Callback>`s?

1c3t3a: Mh... does this make a difference?

ZetaNumbers: Codegen, but nothing else i think.

1c3t3a: Then I would rather leave it as is.

1c3t3a:
I am sorry, I missunderstood the mut thing here, could you drop commit 5800990? Then we just merge the first one.

ZetaNumbers:
Why drop https://github.com/1c3t3a/rust-socketio/commit/5800990eb9aa31e97f1bcbdcab31da14eb933552?

1c3t3a:
> Why drop [5800990](https://github.com/1c3t3a/rust-socketio/commit/5800990eb9aa31e97f1bcbdcab31da14eb933552)?

Currently the builder method doesn't need the mut, but it would be possible (and tbh also expected) that the builder needs mut here. If we were not putting mut here, adding it would be a breaking change, so we might as well just have it there for the case that we ever need mut in a future version of the interface. I didn't realize that acquiring the lock and putting something there doesn't need mut as it uses interior mutability. So I thought we could just drop the `allow(unused_mut)` comment which was wrong.

ZetaNumbers:
> If we were not putting mut here, adding it would be a breaking change

It's nothing like that. This mut isn't even a part of the signature, but the part of the definition. You can mutate immutable value by moving it out of the original immutable variable to the mutable one and mutate it however you want.

1c3t3a:
Mhh fair point because we don't take `&self` here but a moved value. Then it make sense to remove it actually.

PR #293, opened 2023-03-16
done

mars1211:


socketio/src/client/builder.rs
```diff
@@ -42,7 +42,6 @@ pub struct ClientBuilder {
     transport_type: TransportType,
     auth: Option<serde_json::Value>,
     pub(crate) reconnect: bool,
-    pub(crate) reconnect_on_disconnect: bool,
```
1c3t3a, line 45: Why do you remove existing functionality? This does not make sense!

1c3t3a:
Looks like you started to work on an older branch! Please take the latest main state to put your changes on top! I won't merge a PR that is actively removing features!!

PR #294, opened 2023-03-17
Fix wrong event 'connect' to 'open'

ollipal:
[current features](https://docs.rs/rust_socketio/latest/rust_socketio/#current-features) at the main page of the documentation lists the supported features correctly, but the [on method's documentation](https://docs.rs/rust_socketio/latest/rust_socketio/struct.SocketBuilder.html#method.on) seems to have a wrong event 'connect' instead of 'open'.

Code to verify:

```rust
        let socket = SocketBuilder::new("http://localhost:3001")
            .on("open", |_,_| println!("Connected!")) // This triggers, "connect" does not trigger
            .connect()
            .expect("Connection failed");
```

It took me a while to figure this out, I hope this helps someone else!

### DIFF
--- a/socketio/src/asynchronous/client/builder.rs
+++ b/socketio/src/asynchronous/client/builder.rs
@@ -91,7 +91,7 @@ impl ClientBuilder {
     }
 
     /// Registers a new callback for a certain [`crate::event::Event`]. The event could either be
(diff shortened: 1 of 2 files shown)


1c3t3a (approved):
Thanks a lot for spotting this!!

PR #314, opened 2023-04-28
fix: allow awaiting async callbacks

### LINKED ISSUE #313: Asynchronous Client: Cannot use await in callbacks
Hi! Thanks for spending your time working on this library. It's great to see support for an asynchronous client.

## Problem Description

Running the given example for the asynchronous client only works when *not* having an `await` statement within the callback. Code after the first `await` in the callback is not executed.
Thus, calling async functions is not possible.

When debugging, I found out that this is caused by the `Stream` implementation for the async client.

##  Steps To Reproduce

Run this Python server:

```python
import socketio
import eventlet

sio = socketio.Server(async_mode='eventlet')
app = socketio.WSGIApp(sio)

@sio.event
def connect(sid, environ):
    print('Client connected:', sid)

@sio.event
def disconnect(sid):
    print('Client disconnected:', sid)

@sio.event
def foo(sid, data):
    print('foo event received:', data)

if __name__ == '__main__':
    eventlet.wsgi.server(eventlet.listen(('localhost', 4200)), app)
```

Connect with this Rust client and see that there is no output (note that "Connected!" will not be printed):

```rust
use futures_util::FutureExt;
use log::info;
use rust_socketio::{asynchronous::ClientBuilder, Event, TransportType};
use std::time::Duration;
use tokio::{
    signal::{
        self,
        unix::{signal, SignalKind},
    },
    time::sleep,
};

#[tokio::main]
async fn main() {

    info!("Starting");

        .on(Event::Connect, |_payload, _client| {
            {
                async move {
                    sleep(Duration::from_secs(1)).await;
                    info!("Connected!");
                }
                .boxed()
            }
        })
        .connect()
        .await
        .expect("Connection failed");

    tokio::select! {
        _ = signal::ctrl_c() => {
            info!("Received SIGINT, shutting down.");
        },
        _ = sigterm.recv() => {
            info!("Received SIGTERM, shutting down.");
        },
    }

}
```

Here's the corresponding Cargo.toml:

```toml
[package]
name = "client"
version = "0.1.0"
edition = "2021"

[dependencies]
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
rust_socketio = { version = "0.4.1-alpha.1", features = ["async"] }
env_logger = { version = "0.10" }
log = "0.4"
tokio = { version = "1", features = ["macros", "rt", "signal", "time"] }
futures-util = "0.3"
```

It seems like the callback is aborted on the first `await`.

## Possible Solution

Debugging around, I found that in the client's stream implementation, the `ready!` macro is used. This macro only returns the results of futures if they return immediately:

https://github.com/1c3t3a/rust-socketio/blob/10d5a4633c4e88cea2ae795de85daf161193f9fe/socketio/src/asynchronous/client/client.rs#L410-L441

One possible solution I found was to use `futures_util::stream::unfold` in a new function on the `Client` instead of implementing stream (something like that):

```rust
pub(crate) fn stream(&mut self) -> impl Stream<Item = Result<Packet>> + '_ {
    stream::unfold(self.socket.clone(), | mut socket | async {
          let packet: Option<std::result::Result<Packet, Error>> = socket.clone().next().await;
          match packet {
              None => None,
              Some(Err(err)) => {
                  let res = self.callback(&Event::Error, err.to_string()).await;
                  Some((Err(err), socket))
              },
              Some(Ok(packet)) => {
                  self.handle_socketio_packet(&packet).await;
                  Some((Ok(packet), socket))
              }
          }
      })
}
```

I created #314 as potential fix.

felix-gohla:
This change allows for awaiting futures in asynchronous callbacks.

Sadly, I did not manage to keep the `Stream` implementation of the client as this would mean storing the `Unfold` somewhere in the client.
Let me know if you see potential issues with this method. :)

Closes #313

### DIFF
--- a/socketio/src/asynchronous/client/client.rs
+++ b/socketio/src/asynchronous/client/client.rs
@@ -1,6 +1,6 @@
-use std::{collections::HashMap, ops::DerefMut, pin::Pin, sync::Arc, task::Poll};
+use std::{collections::HashMap, ops::DerefMut, pin::Pin, sync::Arc};
 
-use futures_util::{future::BoxFuture, ready, FutureExt, Stream, StreamExt};
+use futures_util::{future::BoxFuture, stream, Stream, StreamExt};
 use log::trace;
 use rand::{thread_rng, Rng};
 use serde_json::{from_str, Value};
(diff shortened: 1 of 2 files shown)


socketio/src/asynchronous/client/client.rs
```diff
@@ -405,51 +405,50 @@ impl Client {
...
-            let next = ready!(self.socket.poll_next_unpin(cx));
-
-            match next {
-                None => {
-                    // end the stream if the underlying one is closed
-                    return Poll::Ready(None);
-                }
+    /// Returns the packet stream for the client.
+    ///
+    /// Note: if multiple
```
1c3t3a, line 411: Did you forget to finish the sentence here?

socketio/src/asynchronous/client/client.rs
```diff
@@ -405,51 +405,50 @@ impl Client {
...
+    pub(crate) fn as_stream<'a>(
+        &'a self,
+    ) -> Pin<Box<dyn Stream<Item = Result<Packet>> + Send + 'a>> {
+        stream::unfold(self.socket.clone(), |mut socket| async {
+            // wait for the next payload
+            let packet: Option<std::result::Result<Packet, Error>> = socket.next().await;
+            println!("{:?}", packet);
```
1c3t3a, line 418: Please remove this line.

socketio/src/asynchronous/client/client.rs
```diff
@@ -516,6 +515,42 @@ mod test {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn socket_io_async_callback() -> Result<()> {
+        // Test whether asynchronous callbacks are fully executed.
+        let url = crate::test::socket_io_server();
+
+        // This synchronization mechanism is used to let the
```
1c3t3a, line 523: nit: Finish or delete this comment please.

socketio/src/asynchronous/client/client.rs
```diff
@@ -516,6 +515,42 @@ mod test {
...
+            .await?;
+
+        let payload = json!({"token": 123_i32});
+        let result = socket
+            .emit("test", Payload::String(payload.to_string()))
+            .await;
+
+        assert!(result.is_ok());
+        // If the timeout did not trigger, the async callback was fully executed.
+        let timeout = timeout(Duration::from_secs(5), notify.notified()).await;
```
1c3t3a, line 548: Does `Notify` have a build in timeout or a timeout option? If yes it might be nicer to use it, otherwise this is good!

1c3t3a (commented):
Thanks a lot for the PR! I have a few comments, but generally this LGTM :)

socketio/src/asynchronous/client/client.rs
felix-gohla: I didn't find one in the [documentation](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html>).

1c3t3a: Alright, then this is the was to go. Thanks for checking :)

PR #324, opened 2023-06-02
fix(async): handle close of packet stream to avoid panic

### LINKED ISSUE #323: Asynchronous Client: Unexpected Server disconnect leads to panic
## Problem Description

In #314 the stream implementation of the asynchronous socket has been adjusted.
However, when the server now disconnects unexpectedly (e.g. process is killed etc.), the following issue can now be observed:

```
thread 'tokio-runtime-worker' panicked at 'Unfold must not be polled after it returned `Poll::Ready(None)`'
```

## Steps To Reproduce

1. Use client and server from #311 
2. Start the server and then the client
3. Stop the server
4. Observe the panic in the client

## Possible Solution

The loop which polls the stream in the asynchronous client builder needs to stop when the stream returns `None`:

https://github.com/1c3t3a/rust-socketio/blob/cb107ba6770fabb109dce173b74ab36ebe039045/socketio/src/asynchronous/client/builder.rs#L346-L355

Apparently, this issue existed before #314 already.
However, due to the different stream implementation, calls to `stream.next()` after the stream has terminated wouldn't panic but instead always return `None` leading to an infinite loop.
The issue only now surfaced because of the different stream implementation.

sirkrypt0:
When the packet stream closes, e.g. because the server terminated the connection, the stream returns None. Then, the next() function shouldn't be called again on the stream, as it is already closed.
Previously, we wouldn't handle this, which lead to a panic.

Before the fix which implemented the as_stream() function for the socket, the program wouldn't panic, but instead run in an endless loop which isn't ideal either.

Now, the loop which fetches the packets from the stream properly terminates. However, the program which uses the socket.io client currently isn't notified of the unexpected closure.

Closes #323 

While I tried writing a test for this, it appears that you can't assert whether some tokio thread panicked or not without having the handle to that thread (there is a feature in `tokio_unstable` called `unhandled_panic` though).
Since the builder currently spawns the stream iterator without returning a handle, I didn't find a way to properly assure the this thread doesn't panic.

### DIFF
--- a/socketio/src/asynchronous/client/builder.rs
+++ b/socketio/src/asynchronous/client/builder.rs
@@ -343,14 +343,10 @@ impl ClientBuilder {
         // Use thread to consume items in iterator in order to call callbacks
         tokio::runtime::Handle::current().spawn(async move {
             let mut stream = socket_clone.as_stream();
-            loop {
-                // tries to restart a poll cycle whenever a 'normal' error occurs,
-                // it just logs on network errors, in case the poll cycle returned
-                // `Result::Ok`, the server receives a close frame so it's safe to
-                // terminate
-                if let Some(e @ Err(Error::IncompleteResponseFromEngineIo(_))) = stream.next().await
-                {
-                    trace!("Network error occured: {}", e.unwrap_err());
+            // Consume the stream until it returns None and the stream is closed.
+            while let Some(item) = stream.next().await {
+                if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = item {
+                    trace!("Network error occurred: {}", e.unwrap_err());
                 }
             }
         });


1c3t3a:
First of all thanks for the PR! I appreciate your contributions to the async version of the crate a lot, the stream impl isn't perfect, but IMO currently the best to do in async rust.

>While I tried writing a test for this, it appears that you can't assert whether some tokio thread panicked or not without having the handle to that thread (there is a feature in tokio_unstable called unhandled_panic though).

Do you think it makes sense to add this feature for tests only? That case we can verify to not brake it and only add a dependency for devs and not for normal users of the crate.

>However, the program which uses the socket.io client currently isn't notified of the unexpected closure.

Yes, and this is an issue (also see the opened issues around it). Do you have an idea how to fix this (e.g. invoke the err callback or something else?). We can't invalidate a `Client` that we already handed out nicely from a background thread anyways, so it is rather about informing the user properly about what is going on.

1c3t3a (approved):

sirkrypt0:
> Do you think it makes sense to add this feature for tests only? That case we can verify to not brake it and only add a dependency for devs and not for normal users of the crate.

I totally agree that it would make sense to test this and tried it out here https://github.com/giz-berlin/rust-socketio/blob/fix/async-stream-close-test/socketio/src/asynchronous/client/client.rs#L553-L600
Sadly, enabling `tokio_unstable` for tests only is rather hacky since it's not a `feature` but a `cfg` and apparently you cannot do the following in `.cargo/config.toml`.

```toml
[target.'cfg(test)']
# Rustflags specifically for tests
rustflags = ["--cfg", "tokio_unstable"]
```

See https://github.com/rust-lang/cargo/issues/8170.
`[target.'cfg(debug_assertions)']` to only enable it on debug builds doesn't seem to work either.

Instead I enabled it via the `RUSTFLAGS` environment variable in the CI...
We may make use of the `Makefile` here though, which is currently not used in the CI I think and set the `RUSTFLAGS` there.
Then, it would at least be consistent for the CI and for users which run `make test...`.

Let me know what you think :)

sirkrypt0:
> Yes, and this is an issue (also see the opened issues around it). Do you have an idea how to fix this (e.g. invoke the err callback or something else?). We can't invalidate a Client that we already handed out nicely from a background thread anyways, so it is rather about informing the user properly about what is going on.

Calling the `error` callback would be possible. However, a limitation I currently see with the callbacks is that they are limited to the `Payload` which is either `Bytes` or a `String`. So determining the actual error type would come down to error-prone string comparison I think.

I think the use of the `error` callback is currently mixed between errors coming from the server and errors occurring locally. For example, the `ConnectError` packet type is handled by calling the `error` callback and concatenating an error message with the actual payload but any local errors are handled by calling the `error` callback as well.

https://github.com/1c3t3a/rust-socketio/blob/39297a54e58cf6e9e4a51f3d5f20167b7d63dc56/socketio/src/client/raw_client.rs#L377-L386

So maybe, the `on` callbacks should be used for events which are sent across the SocketIO connection only and be always passed the packet payload.
Then, for any errors which occur locally (probably all of the `socketio::Error` enum) a separate `.on_error` callback may be used which gets the actual error? Although it is questionable whether users can actually do useful things when they can distinguish the exact error type like `InvalidUtf8` or `InvalidPoisonedLock` (as it says in the TODO: `Do not expose non-trivial internal errors`).

Additionally, it may be useful to have a separate `on_disconnect` callback which is called when a disconnect (either normal closure via an `Event::Close` or unexpected closure) occurs to make the disconnection explicit to the user. This would avoid that users have to parse the error to decide whether an unexpected disconnect occurs?
It could be implemented in a similar fashion to https://socket.io/docs/v4/client-api/#event-disconnect which provides a reason for the disconnect.

1c3t3a:
> Instead I enabled it via the RUSTFLAGS environment variable in the CI...
We may make use of the Makefile here though, which is currently not used in the CI I think and set the RUSTFLAGS there.
Then, it would at least be consistent for the CI and for users which run make test....

+1 This sounds like a workable approach, can you maybe create an issue for when to removing it and add a doc comment? I think it's just good if those tests actually run.

> I think the use of the error callback is currently mixed between errors coming from the server and errors occurring locally. For example, the ConnectError packet type is handled by calling the error callback and concatenating an error message with the actual payload but any local errors are handled by calling the error callback as well.

Yeah this isn't optimal. I think parsing the error is the worst approach to take here, so the error callback shouldn't be the thing to use.

> So maybe, the on callbacks should be used for events which are sent across the SocketIO connection only and be always passed the packet payload.
Then, for any errors which occur locally (probably all of the socketio::Error enum) a separate .on_error callback may be used which gets the actual error?

I agree with you here. Maybe we can split the error enum to errors that we want to expose and internal errors that don't matter to the user but only to us.

Also +1 to the `on_disconnect` callback, sounds reasonable. 

I hope to find time to work on this the upcoming weeks, if not, feel free to pick something up! :) And thanks for the your comments here!

1c3t3a (approved):

PR #325, opened 2023-06-04
perf(engineio): optimize Payload: TryFrom<Bytes>

MaxOhn:
Replacing repeated `Vec::push` with a single `Iterator::collect` which is simpler, more idiomatic and more performant.

The [memchr](https://github.com/BurntSushi/memchr) crate can be used to speed this up further but it's outside the scope of this PR.


1c3t3a (approved):
Nice! Thanks for the PR :)

PR #326, opened 2023-06-04
perf(engineio): optimize headers

MaxOhn:
- Represent `HeaderName` by `Box<str>` instead of `String` to reduce its size.
- Replace `HeaderName: ToString` with `HeaderName: Display` which comes with a free `ToString` impl
- Avoid an allocation when creating `HttpHeaderName `
- Create `HttpHeaderMap` via `Iterator::collect` instead of repeated `HttpHeaderMap::append`

### DIFF
--- a/engineio/src/header.rs
+++ b/engineio/src/header.rs
@@ -6,11 +6,12 @@ use http::{
 };
 use std::collections::HashMap;
 use std::convert::TryFrom;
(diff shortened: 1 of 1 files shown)


1c3t3a (approved):
Good catch! Thanks a lot for the PR!

PR #330, opened 2023-06-04
perf(socketio): avoid allocs in Bytes: From<&Packet>

MaxOhn:
Avoiding intermediate allocations by always writing into the same `String` buffer.

### DIFF
--- a/socketio/src/packet.rs
+++ b/socketio/src/packet.rs
@@ -118,20 +118,16 @@ impl From<&Packet> for Bytes {
         }
 
         if packet.attachments.is_some() {
+            let num = packet.attachment_count - 1;
+
             // check if an event type is present
             if let Some(event_type) = packet.data.as_ref() {
                 let _ = write!(
                     buffer,
-                    "[{},{{\"_placeholder\":true,\"num\":{}}}]",
-                    event_type,
-                    packet.attachment_count - 1,
+                    "[{event_type},{{\"_placeholder\":true,\"num\":{num}}}]",
                 );
             } else {
-                let _ = write!(
-                    buffer,
-                    "[{{\"_placeholder\":true,\"num\":{}}}]",
-                    packet.attachment_count - 1,
-                );
+                let _ = write!(buffer, "[{{\"_placeholder\":true,\"num\":{num}}}]");
             }
         } else if let Some(data) = packet.data.as_ref() {
             buffer.push_str(data);
--- a/socketio/src/packet.rs
+++ b/socketio/src/packet.rs
@@ -1,8 +1,9 @@
 use crate::error::{Error, Result};
 use crate::Error::{InvalidJson, InvalidUtf8};
-use bytes::{BufMut, Bytes, BytesMut};
+use bytes::Bytes;
 use regex::Regex;
 use std::convert::TryFrom;
+use std::fmt::Write;
 
 /// An enumeration of the different `Packet` types in the `socket.io` protocol.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


socketio/src/packet.rs
```diff
@@ -96,50 +97,47 @@ impl From<&Packet> for Bytes {
     /// stream as it gets handled and send by it's own logic via the socket.
     fn from(packet: &Packet) -> Bytes {
         // first the packet type
-        let mut string = (packet.packet_type as u8).to_string();
+        let mut buffer = String::with_capacity(64);
```
1c3t3a, line 100: Can you please add a comment on why you settled for 64 here?

socketio/src/packet.rs
```diff
@@ -96,50 +97,47 @@ impl From<&Packet> for Bytes {
...
-        buffer.put(string.as_ref());
-        if packet.attachments.as_ref().is_some() {
+        if packet.attachments.is_some() {
             // check if an event type is present
-            let placeholder = if let Some(event_type) = packet.data.as_ref() {
-                format!(
+            if let Some(event_type) = packet.data.as_ref() {
+                let _ = write!(
+                    buffer,
                     "[{},{{\"_placeholder\":true,\"num\":{}}}]",
```
1c3t3a, line 125: While you're at it: Can you maybe switch to inline style here as well: `"[{event_type},{{\"_placeholder\":true,\"num\":{packet.attachment_count - 1}}}]"`

socketio/src/packet.rs
```diff
@@ -96,50 +97,47 @@ impl From<&Packet> for Bytes {
...
                     event_type,
                     packet.attachment_count - 1,
-                )
+                );
             } else {
-                format!(
+                let _ = write!(
+                    buffer,
                     "[{{\"_placeholder\":true,\"num\":{}}}]",
```
1c3t3a, line 132: Same as above! :)

socketio/src/packet.rs
```diff
@@ -96,50 +97,47 @@ impl From<&Packet> for Bytes {
...
-                )
-            };
-
-            // build the buffers
-            buffer.put(placeholder.as_ref());
+                );
+            }
         } else if let Some(data) = packet.data.as_ref() {
-            buffer.put(data.as_ref());
+            buffer.push_str(data);
```
1c3t3a, line 137: Can't we use `write!` as with the id?

1c3t3a (commented):
Hi, thanks a lot for the PR, very much appreciated! :) I left a few comments, but generally this LGTM!

socketio/src/packet.rs
MaxOhn: Good point, I forgot that there's not necessarily attachments and allocated for them either way. Probably wiser to not pre-allocate all.

socketio/src/packet.rs
MaxOhn: Currently rust only allows you to put variables themselves into `{}` so I can't add the expression directly but putting the subtraction into its own variable beforehand would work. I agree that'd look better 😃

socketio/src/packet.rs
MaxOhn: We could yea. Whereas for id I had no choice but to use the macro because the integer needed to be stringified, generally the macro should be avoided if possible when looking for performance since it might bring in some unnecessary indirections or boilerplate code under-the-hood.

1c3t3a: Fair, although I would trust the compiler to not emit stuff we don't need here. But you have a point, lets keep it as is :)

1c3t3a (approved):
LGTM! Thanks. Feel free to merge once you're ready.

socketio/src/packet.rs
1c3t3a: Agreed, lets hope we get support for any expressions soon :)

PR #331, opened 2023-06-05
perf(socketio): optimize Packet: TryFrom<&Bytes>

MaxOhn:
Rewrite of the `Packet: TryFrom<&Bytes>` implementation.

I ran a benchmark on randomly generated packets and saw a performance improvement of up to 15x 😮
Can push the benchmark to a public repository if you want to verify.

This also gets rid of the direct dependency on `regex`.

Haven't touched the test coverage since it seems decent already. If you want I can try figuring out and adding some more test cases.

### DIFF
--- a/socketio/src/client/raw_client.rs
+++ b/socketio/src/client/raw_client.rs
@@ -270,36 +270,32 @@ impl RawClient {
     /// Handles the incoming acks and classifies what callbacks to call and how.
     #[inline]
     fn handle_ack(&self, socket_packet: &Packet) -> Result<()> {
-        let mut to_be_removed = Vec::new();
-        if let Some(id) = socket_packet.id {
-            for (index, ack) in self.outstanding_acks.lock()?.iter_mut().enumerate() {
-                if ack.id == id {
-                    to_be_removed.push(index);
+        let Some(id) = socket_packet.id else {
+            return Ok(());
+        };
 
-                    if ack.time_started.elapsed() < ack.timeout {
-                        if let Some(ref payload) = socket_packet.data {
-                            ack.callback.deref_mut()(
-                                Payload::String(payload.to_owned()),
-                                self.clone(),
-                            );
-                        }
-                        if let Some(ref attachments) = socket_packet.attachments {
-                            if let Some(payload) = attachments.get(0) {
-                                ack.callback.deref_mut()(
-                                    Payload::Binary(payload.to_owned()),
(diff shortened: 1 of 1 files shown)


socketio/src/packet.rs
```diff
@@ -154,102 +153,61 @@ impl TryFrom<&Bytes> for Packet {
...
-        let mut utf8_iter = payload_utf8.chars().peekable();
-        let mut next_utf8;
-        let mut char_buf: Vec<char> = vec![];
+        let mut payload = str_from_utf8(&payload).map_err(Error::InvalidUtf8)?;
+        let mut packet = Packet::default();
 
         // packet_type
-        packet.packet_type =
-            PacketId::try_from(utf8_iter.next().ok_or(Error::IncompletePacket())?)?;
+        let c = payload.chars().next().ok_or(Error::IncompletePacket())?;
```
1c3t3a, line 160: nit: Can you please pick a more descriptive name than 'c'.

1c3t3a (approved):
LGTM! Great, really happy to hear that about 15x perf improvement, very well appreciated :) 

Just out of curiousity, what part of the code do you think makes the difference?

1c3t3a:
Can you please drop the merge commit and instead rebase the other changes? :)

MaxOhn:
> LGTM! Great, really happy to hear that about 15x perf improvement, very well appreciated :)
> 
> Just out of curiousity, what part of the code do you think makes the difference?

It's a combination of multiple things I would say.
- The `json_data` was not only validated but also used to create the string. `serde_json::Value::to_string` is pretty costly and wasn't even necessary because the json string it was built with could have been used instead.
- `.to_vec()`, `.to_string()`, `.collect()`, `.to_owned()`, and even `str::replace` each come with an allocation, never a good sign.
- Building a `Regex` at runtime is pretty expensive. As luck would have it, the most recent rust version 1.70 introduced [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) which would have allowed to compile it once and re-use it but the better alternative of course is to not use regex at all in the first place.
- I replaced a bunch of manual loops and their branching. Iterator methods are often times faster since rust is so fine-tuned for those. And using std lib methods wherever applicable in general is likely a good idea, they consist of performant magic. 😄

MaxOhn:
Oh apologies, didn't mean to close / merge this just yet with the force push after rebasing 😅

PR #332, opened 2023-06-05
perf(socketio): replace Vec::remove with Vec::retain

MaxOhn:
Replacing manual removal via intermediate Vec and `Vec::remove` with a more performant `Vec::retain_mut`.


1c3t3a (approved):
LGTM! Thanks :)

PR #334, opened 2023-06-05
perf(socketio): avoid allocs when dealing with events

MaxOhn:
This avoids some more allocations by
- formatting events directly instead of using `String: From<Event>`
- creating events directly instead of using `Event: From<&str>`

I also copied the `handle_event` method of the async client for the non-async client.


MaxOhn:
Another change I considered doing was to replace `Event::Custom(String)` with `Event::Custom(Box<str>)` which would reduce the size of `Event` by 25% from 32 bytes to 24 bytes.

Since `Event` is a public enum, this would be a breaking change so I didn't include it yet.
Thoughts on this?

1c3t3a (approved):
LGTM! Thank you :) About changing `Payload::String`, as you also said I think it is not necessary to have that kind of performance improvement. And it would introduce a breaking change which is also unnecessary.

PR #357, opened 2023-09-10
Add async client support

ehuff700:
I'm newer to Rust so I would definitely appreciate a code review here. I believe I've got it successfully working and it seems to pass the tests that I was able to run. I'm using this fork right now for my really basic socket.io client/server implementation and it's working quite well.

### DIFF
--- a/socketio/Cargo.toml
+++ b/socketio/Cargo.toml
@@ -7,7 +7,11 @@ description = "An implementation of a socketio client written in rust."
 readme = "../README.md"
 repository = "https://github.com/1c3t3a/rust-socketio"
 keywords = ["socketio", "engineio", "network", "protocol", "client"]
-categories = ["network-programming", "web-programming", "web-programming::websocket"]
+categories = [
+    "network-programming",
+    "web-programming",
+    "web-programming::websocket",
+]
 license = "MIT"
 
 [dependencies]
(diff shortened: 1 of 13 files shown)


socketio/Cargo.toml
```diff
@@ -7,7 +7,11 @@ description = "An implementation of a socketio client written in rust."
 readme = "../README.md"
 repository = "https://github.com/1c3t3a/rust-socketio"
 keywords = ["socketio", "engineio", "network", "protocol", "client"]
-categories = ["network-programming", "web-programming", "web-programming::websocket"]
+categories = [
+    "network-programming",
+    "web-programming",
+    "web-programming::websocket",
+]
 license = "MIT"
```
1c3t3a, line 15: Is there any good reason you change the formatting in this file? If you foilow a special linter would be nice to know which one and maybe include a config? Or don't reformat?

socketio/src/asynchronous/client/builder.rs
```diff
@@ -1,34 +1,44 @@
...
     opening_headers: Option<HeaderMap>,
     transport_type: TransportType,
     auth: Option<serde_json::Value>,
+    pub(crate) reconnect: bool,
+    pub(crate) reconnect_on_disconnect: bool,
+    // None reconnect attempts represent infinity.
+    pub(crate) max_reconnect_attempts: Option<u8>,
+    pub(crate) reconnect_delay_min: u64,
+    pub(crate) reconnect_delay_max: u64,
 }
```
1c3t3a, line 42: How comes you're moving all logic to the builder? You're inverting Client and Builder, the Builder has all logic and holds a Client, while the Client just has a builder. That doesn't really make sense.

1c3t3a (commented):
Hey there, first of all sorry for the late review! I am not sure about the purpose of this PR - this crate already has async support (behind a flag). In the code you're mostly switching the Client and Builder. Can you maybe elaborate about the purpose of this PR? If you want we could also chat directly, just shoot me an email.

1c3t3a:
Closing this for now.

PR #379, opened 2023-11-30
add Payload::StringArray

SalahaldinBilal:
adding variant StringArray for Payload to support sending multiple messages and also to send empty messages (by using empty array)

### DIFF
--- a/socketio/src/asynchronous/socket.rs
+++ b/socketio/src/asynchronous/socket.rs
@@ -115,9 +115,33 @@ impl Socket {
                 Some(vec![bin_data]),
             )),
             Payload::String(str_data) => {
+                // is there any reason this is different than the blocking client ?
                 serde_json::from_str::<IgnoredAny>(&str_data)?;
 
                 let payload = format!("[\"{event}\",{str_data}]");
+
+                Ok(Packet::new(
+                    PacketId::Event,
+                    nsp.to_owned(),
+                    Some(payload),
+                    id,
+                    0,
+                    None,
+                ))
+            }
+            Payload::StringArray(array) => {
+                let stringified_array = array
+                    .iter()
+                    .map(|string| {
+                        if serde_json::from_str::<IgnoredAny>(&string).is_ok() {
+                            string.to_owned()
+                        } else {
+                            format!("\"{string}\"")
+                        }
+                    })
+                    .fold(String::new(), |acc, current| format!("{acc}, {current}"));
+
+                let payload = format!("[\"{event}\"{stringified_array}]");
 
                 Ok(Packet::new(
                     PacketId::Event,
--- a/socketio/src/payload.rs
+++ b/socketio/src/payload.rs
@@ -2,18 +2,26 @@ use bytes::Bytes;
 
 /// A type which represents a `payload` in the `socket.io` context.
 /// A payload could either be of the type `Payload::Binary`, which holds
-/// data in the [`Bytes`] type that represents the payload or of the type
-/// `Payload::String` which holds a [`std::string::String`]. The enum is
+/// data in the [`Bytes`] type that represents the payload, of the type
+/// `Payload::String` which holds a [`std::string::String`] or of the type
+/// `Payload::StringArray` which holds a [`Vec<std::string::String>`]. The enum is
 /// used for both representing data that's send and data that's received.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum Payload {
     Binary(Bytes),
     String(String),
+    StringArray(Vec<String>),
 }
 
 impl From<&str> for Payload {
     fn from(string: &str) -> Self {
         Self::String(string.to_owned())
+    }
+}
+
+impl From<Vec<&str>> for Payload {
+    fn from(strings: Vec<&str>) -> Self {
+        Self::StringArray(strings.iter().map(|str| str.to_string()).collect())
     }
 }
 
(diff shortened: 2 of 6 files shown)


SalahaldinBilal:
Hey @1c3t3a, thanks for the reply, while an array string payload _can_ get the job done, it is not the same, since it sends all the elements in the array as a single **argument** in the message, while the JS socket.io server and client can respectively recieve and send **multiple arguments**, i made an example tauri app that has both this client and the JS client to highlight the difference, you can find it [here](https://github.com/SalahaldinBilal/socket-io-params-example).

If you don't wanna bother with running the app, here's the output on the server side, you can clearly see how the rust client sent only a single argument, but the JS client sent multiple agruments, and that's the point of this PR, to allow that behaivor to exist also in this client.

WindowsTerminal_GneCc8zHcb

1c3t3a:
Thanks for the explanation, that sounds reasonable. Thanks for the fix as well!

Unfortunately this introduces a major breaking change and might break users of this crate.

I would suggest to add this behind a feature (e.g. string_array), that way people can upgrade rust_socketio without breaking their existing code.

Furthermore for landing this without a feature I would start discussing whether it makes sense to switch the payload entirely to return a `serde_json::Json` instead of a `String`.

1c3t3a:
@ctrlaltf24 What do you think of that? Would be great to get your opinion :)

socketio/src/asynchronous/socket.rs
```diff
@@ -115,10 +115,34 @@ impl Socket {
                 Some(vec![bin_data]),
             )),
             Payload::String(str_data) => {
+                // is there any reason this is different than the blocking client ?
                 serde_json::from_str::<IgnoredAny>(&str_data)?;
```
1c3t3a, line 119: Good point! Could you correct it please? Should be the same as the synchronous client.

socketio/src/payload.rs
```diff
@@ -52,6 +72,8 @@ impl AsRef<[u8]> for Payload {
         match self {
             Payload::Binary(b) => b.as_ref(),
             Payload::String(s) => s.as_ref(),
+            // TODO: how to handle this case
+            Payload::StringArray(_a) => &[0u8; 0],
```
1c3t3a, line 76: Mh... sounds like we would need to drop this impl. I don't see any way this would make sense...

1c3t3a (commented):
Is it maybe possible to extend the socket.io server test? That way we can make sure to not accidentally break this in the future.

Also it looks like one of the examples is broken.

1c3t3a:
If you need help with adding a feature, just ask! :)

socketio/src/asynchronous/socket.rs
ctrlaltf24: Perhaps move to a pub(crate) fn on Packet?

socketio/src/payload.rs
```diff
@@ -2,13 +2,15 @@ use bytes::Bytes;
...
-/// `Payload::String` which holds a [`std::string::String`]. The enum is
+/// data in the [`Bytes`] type that represents the payload, of the type
+/// `Payload::String` which holds a [`std::string::String`] or of the type
+/// `Payload::StringArray` which holds a [`Vec<std::string::String>`]. The enum is
 /// used for both representing data that's send and data that's received.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum Payload {
     Binary(Bytes),
     String(String),
+    StringArray(Vec<String>),
```
ctrlaltf24, line 13: (reflecting on previous decision to use string and conditionally interoperate as JSON) Is string really the type we want here? From our usage (checking to see if the string is valid JSON) it looks more like we mean serde_json::Value.

We could use an intermediate type with some From conversions to make the ergonomic

```rust
struct TextPayload(serde_json::Value);

enum Payload {
  Binary(Vec<u8>),
  Text(Vec<TextPayload>)
}

impl From<String> for Payload {
  fn from(string: String) -> Self {
    Self::Text(vec![TextPayload::from(string)])
  }
}

impl From<String> for TextPayload {
  fn from(string: String) -> Self {
    if let Ok(value) = serde_json::from_str::<serde_json::Value>(&string) {
        Self(value)
    } else {
        Self(serde_json::Value::String(string))
    }
  }
}
```

ctrlaltf24: Supporting multiple arguments seems like a reasonable reason for a minor version bump IMO. Everyone will want this feature

ctrlaltf24 (approved):
Seems reasonable, will need to do a minor version bump, but good find!

Not directly related to your changes, but had an idea to consider using serde_json::Value instead of String for paylod

socketio/src/payload.rs
SalahaldinBilal: this sounds great, and storing the serde value directly will definitly be easier, i'll get working on it when i get some free time

SalahaldinBilal: but wouldn't there be problems with existing code if the enum variant name and inner value changed ?

socketio/src/payload.rs
SalahaldinBilal: yeah was kinda confused where this would be used, and dropping it will make life easier

socketio/src/payload.rs
ctrlaltf24: > but wouldn't there be problems with existing code if the enum variant name and inner value changed ?

Yes, fair point. We should be nice to existing users and deprecate the old one. Updated original post to include the current String variant.

existing `match` expressions on this will also break with the addition of any variants to Payload, which seems unavoidable in order to support this feature (which we should).

SalahaldinBilal:
Feature added in #384

PR #310, opened 2023-04-16
Implemented a response to the server after the client received the message.

mars1211:
Implemented the functionality from my question: https://github.com/1c3t3a/rust-socketio/issues/292

PR #384, opened 2023-12-05
feat!: support multiple arguments to the payload

ctrlaltf24:
Added new payload variant called `Text` to support multiple arguments encoded in JSON. The old Payload::String is deprecated, and will be removed shortly. Use `Payload::from(String)` to mimic old behavior

Built on top of @SalahaldinBilal 's work https://github.com/1c3t3a/rust-socketio/pull/379

Also fixed running doc tests in the devcontainer

Issue #95, opened 2021-08-25
[Help] Respond to incoming event with a result payload?


namxam:
My server emits an event and is expecting a return value from the receiving client. I see there is `emit_with_ack()`, but can I somehow send a payload when processing an incoming event?

1c3t3a:
Yes, looking at the [function signature](https://docs.rs/rust_socketio/0.2.4/rust_socketio/struct.Socket.html#method.emit_with_ack), the parameter `event` takes an `FnMut(Payload, Socket)`. The `Payload` is the payload send by the server and the `Socket` is an instance of the socket. That way you could use the socket inside the Callback, e.g. like that:

```rust
let mut socket = SocketBuilder::new("http://localhost:4200")
    .on("foo", |payload: Payload, _| println!("Received: {:#?}", payload))
    .connect()
    .expect("connection failed");



let ack_callback = |message: Payload, my_socket: Socket| {
    match message {
        Payload::String("send me a message!!!") => my_socket.emit("foo", "answer is 42".into()),
        Payload::String(s) => println!("{}", s),
        _ => (),
   }    
};

let payload = json!({"token": 123});
socket.emit_with_ack("foo", payload, Duration::from_secs(2), ack_callback).unwrap();
```

1c3t3a:
Does this solve your issue? Or did you mean something else?

PR #385, opened 2023-12-11
Emit answer rework

shenjackyuanjie:
a rework of #310 
WARNING: this PR is not fully tested

please be free to add any commit to this PR

### DIFF
--- a/socketio/src/asynchronous/socket.rs
+++ b/socketio/src/asynchronous/socket.rs
@@ -84,9 +84,59 @@ impl Socket {
     /// Emits to certain event with given data. The data needs to be JSON,
     /// otherwise this returns an `InvalidJson` error.
     pub async fn emit(&self, nsp: &str, event: Event, data: Payload) -> Result<()> {
-        let socket_packet = Packet::new_from_payload(data, event, nsp, None)?;
+        let socket_packet = self.build_packet_for_payload(data, event, nsp, None, false)?;
 
         self.send(socket_packet).await
+    }
+
+    /// Returns a packet for a payload, could be used for bot binary and non binary
+    /// events and acks. Convenance method.
+    #[inline]
+    pub(crate) fn build_packet_for_payload<'a>(
+        &'a self,
+        payload: Payload,
+        event: Event,
+        nsp: &'a str,
+        id: Option<i32>,
+        is_answer: bool,
+    ) -> Result<Packet> {
+        match payload {
+            Payload::Binary(bin_data) => Ok(Packet::new(
+                if id.is_some() {
+                    PacketId::BinaryAck
+                } else {
+                    PacketId::BinaryEvent
+                },
+                nsp.to_owned(),
+                Some(serde_json::Value::String(event.into()).to_string()),
+                id,
+                1,
+                Some(vec![bin_data]),
+            )),
+            Payload::String(str_data) => {
+                serde_json::from_str::<serde_json::Value>(&str_data)?;
+
+                let package_type;
+                let payload;
+                if is_answer {
+                    payload = format!("[{}]", str_data);
+                    package_type = PacketId::Ack;
+                } else {
+                    payload = format!("[\"{}\",{}]", String::from(event), str_data);
+                    package_type = PacketId::Event;
+                }
+
+                Ok(Packet::new(
+                    package_type,
+                    nsp.to_owned(),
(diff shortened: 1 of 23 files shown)


socketio/src/asynchronous/socket.rs
```diff
@@ -84,11 +84,61 @@ impl Socket {
...
+                Ok(Packet::new(
+                    package_type,
+                    nsp.to_owned(),
+                    Some(payload),
+                    id,
+                    0,
+                    None,
+                ))
+            },
+            Payload::Text(data) => todo!("not implemented yet"),
```
shenjackyuanjie, line 138: first todo

socketio/src/socket.rs
```diff
@@ -73,11 +73,68 @@ impl Socket {
...
+                Ok(Packet::new(
+                    package_type,
+                    nsp.to_owned(),
+                    Some(payload),
+                    id,
+                    0,
+                    None,
+                ))
+            },
+            Payload::Text(data) => todo!("Text: {:?}", data),
```
shenjackyuanjie, line 134: second todo

shenjackyuanjie (commented):
the two todo I don't know how to do it

shenjackyuanjie:
there is still some conflict found with #384

shenjackyuanjie:
@1c3t3a  hello?(

1c3t3a:
Hey @shenjackyuanjie, I'll try and get to it over the holidays!

shenjackyuanjie:
> Hey @shenjackyuanjie, I'll try and get to it over the holidays!

got it! have a nice day~

shenjackyuanjie:
and I think I should fix the test first(

shenjackyuanjie:
I just remove the new test, which will cause issue, and can't test any new feature(

shenjackyuanjie:
@1c3t3a  if is available, I think you can remove the workflows approve(
since I can't run test locally

1c3t3a:
Seems like this is still failing a bunch of doc tests.

1c3t3a:
Actually I am not very happy with the implementation here. As stated in the original issue, I am not sure whether this is really needed: (see #95). Also this starts to expose the Ack ID to *every* callback we ever define and reworks *all* our public callback APIs. Until now, the Ack ID is never exposed and an implementation detail that is not exposed to the outside world. If we would ever want to expose this, I would rather introduce a new new Callback type that allows for that functionality and pass that to selected public APIs.

In the current state this is changing all public APIs, which I don't think is reasonable for such a small feature.

Totodore:
I'm not really familiar to this codebase but I have another proposition, because the `ack` id is related to the received payload. Why can't we create an `AckSender` struct that is part of the payload and that contains a `Clone` of `InnerSocket` and the ack id.
The `AckSender` has an `emit` method to "answer" with an `acknowledgement` corresponding to the ack id. 

```rust
pub enum Payload {
    Binary(Bytes),
    String(String),
    StringAck(String, AckSender),
    BinaryAck(Bytes, AckSender),
}
```
It is basically what I did here for the [socketio server implementation](https://github.com/Totodore/socketioxide/blob/39bc2f9d8daad4ad8f7cbe411df6a24ca35f4124/socketioxide/src/handler/extract.rs#L247-L254) : 
```rust
/// An Extractor to send an ack response corresponding to the current event.
/// If the client sent a normal message without expecting an ack, the ack callback will do nothing.
#[derive(Debug)]
pub struct AckSender<A: Adapter = LocalAdapter> {
    binary: Vec<Vec<u8>>,
    socket: Arc<Socket<A>>,
    ack_id: Option<i64>,
}
...
pub fn send(self, data: impl Serialize) -> Result<(), SendError> {
    if let Some(ack_id) = self.ack_id {
        let ns = self.socket.ns();
        let data = serde_json::to_value(&data)?;
        let packet = if self.binary.is_empty() {
            Packet::ack(ns, data, ack_id)
        } else {
            Packet::bin_ack(ns, data, self.binary, ack_id)
        };
        self.socket.send(packet)
    } else {
        Ok(())
    }
}
```

Whatever the way it is implemented, it would be really helpful to have this feature :)

shenjackyuanjie:
@1c3t3a how about I spilt this PR into two
or just remove the Ack part?
for now, master branch still can't receive multiple payload at once

PR #392, opened 2024-01-24
Fix the implementation for the multi payload

shenjackyuanjie:
tested on true workflow

a part of #385

and will fix #409

PR #394, opened 2024-02-01
Let handle_event provide Vec<Value> to self.callback

kitgxrl:
This allows the client to receive multiple messages in one event if a server sends that way, e.g. `42["chat", user_information, chat_message]` is one of many events in my server I'm working with at the moment. Some of this can be chalked up to poor socket design but it's still something socket.io allows.

Unsure if this is the best way to approach this problem! It appears to work in my project fine with little to no issues.

### DIFF
--- a/socketio/src/asynchronous/client/client.rs
+++ b/socketio/src/asynchronous/client/client.rs
@@ -328,7 +328,7 @@ impl Client {
         };
 
         // a socketio message always comes in one of the following two flavors (both JSON):
-        // 1: `["event", "msg"]`
+        // 1: `["event", "msg", ...]`
         // 2: `["msg"]`
         // in case 2, the message is ment for the default message event, in case 1 the event
         // is specified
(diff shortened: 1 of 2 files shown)


shenjackyuanjie:
@kitgxrl 
have a look at my PR, #392 
it's the same thing

kitgxrl:
> @kitgxrl have a look at my PR, #392 it's the same thing

Sorry! Only now noticed that issue was the same thing, thought it was regarding something else! That's my mistake. Closing now.

shenjackyuanjie:
it's fine, I have notice that you modify the async client and the `raw_client` would you like to test my PR since I only modify `raw_client`, async part may not work

kitgxrl:
Hey! Sorry for late reply. Been busy with personal stuff. I personally don't use async client since the lack of reconnection is kinda a deal breaker since I need to be able to reliably reconnect.

PR #400, opened 2024-03-18
Async reconnections

### LINKED ISSUE #291: Async ClientBuilder dose not impl reconnect
I use `tokio` and need `reconnect` option for `async` ClientBuilder but blocking ClientBuilder have one.

Can implement reconnect for `async` version?

rageshkrishna:
The async client and builder can now attempt to reconnect to the server on connection errors.

Fixes #291

### DIFF
--- a/socketio/src/asynchronous/client/builder.rs
+++ b/socketio/src/asynchronous/client/builder.rs
@@ -1,4 +1,4 @@
-use futures_util::{future::BoxFuture, StreamExt};
+use futures_util::future::BoxFuture;
 use log::trace;
 use native_tls::TlsConnector;
 use rust_engineio::{
@@ -8,7 +8,7 @@ use rust_engineio::{
 use std::collections::HashMap;
 use url::Url;
 
-use crate::{error::Result, Error, Event, Payload, TransportType};
+use crate::{error::Result, Event, Payload, TransportType};
 
 use super::{
     callback::{Callback, DynAsyncAnyCallback, DynAsyncCallback},
@@ -22,13 +22,13 @@ use crate::asynchronous::socket::Socket as InnerSocket;
 /// acts the `build` method and returns a connected [`Client`].
 pub struct ClientBuilder {
     address: String,
-    on: HashMap<Event, Callback<DynAsyncCallback>>,
-    on_any: Option<Callback<DynAsyncAnyCallback>>,
-    namespace: String,
+    pub(crate) on: HashMap<Event, Callback<DynAsyncCallback>>,
+    pub(crate) on_any: Option<Callback<DynAsyncAnyCallback>>,
+    pub(crate) namespace: String,
     tls_config: Option<TlsConnector>,
     opening_headers: Option<HeaderMap>,
     transport_type: TransportType,
-    auth: Option<serde_json::Value>,
+    pub(crate) auth: Option<serde_json::Value>,
     pub(crate) reconnect: bool,
     pub(crate) reconnect_on_disconnect: bool,
     // None implies infinite attempts
@@ -383,26 +383,14 @@ impl ClientBuilder {
     /// }
     /// ```
     pub async fn connect(self) -> Result<Client> {
-        let socket = self.connect_manual().await?;
-        let socket_clone = socket.clone();
-
-        // Use thread to consume items in iterator in order to call callbacks
-        tokio::runtime::Handle::current().spawn(async move {
-            let mut stream = socket_clone.as_stream();
-            // Consume the stream until it returns None and the stream is closed.
-            while let Some(item) = stream.next().await {
-                if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = item {
-                    trace!("Network error occurred: {}", e.unwrap_err());
-                }
-            }
-        });
+        let mut socket = self.connect_manual().await?;
+        socket.poll_stream().await?;
 
         Ok(socket)
     }
 
-    //TODO: 0.3.X stabilize
-    pub(crate) async fn connect_manual(self) -> Result<Client> {
-        // Parse url here rather than in new to keep new returning Self.
+    /// Creates a new Socket that can be used for reconnections
+    pub(crate) async fn inner_create(&self) -> Result<InnerSocket> {
         let mut url = Url::parse(&self.address)?;
 
         if url.path() == "/" {
@@ -411,11 +399,11 @@ impl ClientBuilder {
 
         let mut builder = EngineIoClientBuilder::new(url);
 
-        if let Some(tls_config) = self.tls_config {
-            builder = builder.tls_config(tls_config);
+        if let Some(tls_config) = &self.tls_config {
+            builder = builder.tls_config(tls_config.to_owned());
         }
-        if let Some(headers) = self.opening_headers {
-            builder = builder.headers(headers);
+        if let Some(headers) = &self.opening_headers {
+            builder = builder.headers(headers.to_owned());
         }
 
         let engine_client = match self.transport_type {
@@ -426,14 +414,14 @@ impl ClientBuilder {
         };
 
         let inner_socket = InnerSocket::new(engine_client)?;
+        Ok(inner_socket)
+    }
 
-        let socket = Client::new(
-            inner_socket,
-            &self.namespace,
-            self.on,
-            self.on_any,
-            self.auth,
-        )?;
+    //TODO: 0.3.X stabilize
+    pub(crate) async fn connect_manual(self) -> Result<Client> {
+        let inner_socket = self.inner_create().await?;
+
+        let socket = Client::new(inner_socket, self)?;
         socket.connect().await?;
 
         Ok(socket)
--- a/socketio/src/asynchronous/client/builder.rs
+++ b/socketio/src/asynchronous/client/builder.rs
@@ -29,6 +29,12 @@ pub struct ClientBuilder {
     opening_headers: Option<HeaderMap>,
     transport_type: TransportType,
     auth: Option<serde_json::Value>,
+    pub(crate) reconnect: bool,
+    pub(crate) reconnect_on_disconnect: bool,
+    // None implies infinite attempts
+    pub(crate) max_reconnect_attempts: Option<u8>,
+    pub(crate) reconnect_delay_min: u64,
+    pub(crate) reconnect_delay_max: u64,
 }
 
 impl ClientBuilder {
@@ -81,6 +87,12 @@ impl ClientBuilder {
             opening_headers: None,
             transport_type: TransportType::Any,
             auth: None,
+            reconnect: true,
+            reconnect_on_disconnect: false,
+            // None implies infinite attempts
+            max_reconnect_attempts: None,
+            reconnect_delay_min: 1000,
+            reconnect_delay_max: 5000,
         }
     }
 
@@ -312,6 +324,34 @@ impl ClientBuilder {
     pub fn transport_type(mut self, transport_type: TransportType) -> Self {
         self.transport_type = transport_type;
 
+        self
+    }
+
+    /// If set to `false` do not try to reconnect on network errors. Defaults to
+    /// `true`
+    pub fn reconnect(mut self, reconnect: bool) -> Self {
+        self.reconnect = reconnect;
+        self
+    }
+
+    /// If set to `true` try to reconnect when the server disconnects the
+    /// client. Defaults to `false`
+    pub fn reconnect_on_disconnect(mut self, reconnect_on_disconnect: bool) -> Self {
+        self.reconnect_on_disconnect = reconnect_on_disconnect;
+        self
+    }
+
+    /// Sets the minimum and maximum delay between reconnection attempts
+    pub fn reconnect_delay(mut self, min: u64, max: u64) -> Self {
+        self.reconnect_delay_min = min;
+        self.reconnect_delay_max = max;
+        self
+    }
+
+    /// Sets the maximum number of times to attempt reconnections. Defaults to
+    /// an infinite number of attempts
+    pub fn max_reconnect_attempts(mut self, reconnect_attempts: u8) -> Self {
+        self.max_reconnect_attempts = Some(reconnect_attempts);
         self
     }
 
--- a/socketio/src/asynchronous/client/client.rs
+++ b/socketio/src/asynchronous/client/client.rs
@@ -29,6 +29,17 @@ use crate::{
     Event, Payload,
 };
 
+#[derive(Default)]
+enum DisconnectReason {
+    /// There is no known reason for the disconnect; likely a network error
+    #[default]
+    Unknown,
+    /// The user disconnected manually
+    Manual,
+    /// The server disconnected
+    Server,
+}
+
 /// A socket which handles communication with the server. It's initialized with
 /// a specific address as well as an optional namespace to connect to. If `None`
 /// is given the client will connect to the default namespace `"/"`.
@@ -42,8 +53,7 @@ pub struct Client {
     // Data send in the opening packet (commonly used as for auth)
     auth: Option<serde_json::Value>,
     builder: Arc<RwLock<ClientBuilder>>,
-    manually_disconnected: Arc<AtomicBool>,
-    server_disconnected: Arc<AtomicBool>,
+    disconnect_reason: Arc<RwLock<DisconnectReason>>,
 }
 
 impl Client {
@@ -58,8 +68,7 @@ impl Client {
             outstanding_acks: Arc::new(RwLock::new(Vec::new())),
             auth: builder.auth.clone(),
             builder: Arc::new(RwLock::new(builder)),
-            manually_disconnected: Arc::new(AtomicBool::new(false)),
-            server_disconnected: Arc::new(AtomicBool::new(false)),
+            disconnect_reason: Arc::new(RwLock::new(DisconnectReason::default())),
         })
     }
 
@@ -85,6 +94,9 @@ impl Client {
         // New inner socket that can be connected
         let mut client_socket = self.socket.write().await;
         *client_socket = socket;
+
+        // Now that we have replaced `self.socket`, we drop the write lock
+        // because the `connect` method we call below will need to use it
         drop(client_socket);
 
         self.connect().await?;
@@ -98,6 +110,8 @@ impl Client {
         let reconnect_delay_min = builder.reconnect_delay_min;
         let reconnect_delay_max = builder.reconnect_delay_max;
         let max_reconnect_attempts = builder.max_reconnect_attempts;
+        let reconnect = builder.reconnect;
+        let reconnect_on_disconnect = builder.reconnect_on_disconnect;
         drop(builder);
 
         let mut client_clone = self.clone();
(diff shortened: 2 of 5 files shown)


socketio/src/asynchronous/client/builder.rs
```diff
@@ -343,26 +385,70 @@ impl ClientBuilder {
...
-            let mut stream = socket_clone.as_stream();
-            // Consume the stream until it returns None and the stream is closed.
-            while let Some(item) = stream.next().await {
-                if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = item {
-                    trace!("Network error occurred: {}", e.unwrap_err());
+            loop {
+                let mut stream = socket_clone.as_stream();
+                // Consume the stream until it returns None and the stream is closed.
+                while let Some(item) = stream.next().await {
+                    // Where does this get thrown by the async client? :confused:
```
1c3t3a, line 401: Ohh good catch! For the Async part, that would either be: 

- `IncompleteResponseFromReqwest`: Thrown for async polling transport defined [here](https://github.com/1c3t3a/rust-socketio/blob/57bcef7ad048d2e72e165ebb6d082dd2bf996ecf/engineio/src/asynchronous/async_transports/polling.rs)
- `WebsocketError`: Thrown for async Websocket transport defined [here](https://github.com/1c3t3a/rust-socketio/blob/57bcef7ad048d2e72e165ebb6d082dd2bf996ecf/engineio/src/asynchronous/async_transports/websocket_general.rs).
-

socketio/src/asynchronous/client/builder.rs
```diff
@@ -343,26 +385,70 @@ impl ClientBuilder {
...
+                    let mut backoff = ExponentialBackoffBuilder::new()
+                        .with_initial_interval(Duration::from_millis(reconnect_delay_min))
+                        .with_max_interval(Duration::from_millis(reconnect_delay_max))
+                        .build();
+
+                    loop {
+                        if let Some(max_reconnect_attempts) = max_reconnect_attempts {
+                            reconnect_attempts += 1;
+                            if reconnect_attempts > max_reconnect_attempts {
+                                // TODO: How to tell the app about this?
```
1c3t3a, line 421: Logging is our best resort at the moment, since there is no way of telling this through a `Result` value at the moment.... Same as with the network error above.

socketio/src/asynchronous/client/builder.rs
```diff
@@ -343,26 +385,70 @@ impl ClientBuilder {
     /// }
     /// ```
     pub async fn connect(self) -> Result<Client> {
```
1c3t3a, line 387: Can be in a follow-up, since you're anyways doing multiple changes: Can this move to a seperate method in the Client? Similar to how we do it for sync transport: https://github.com/1c3t3a/rust-socketio/blob/main/socketio/src/client/client.rs#L212-L243.

socketio/src/asynchronous/client/client.rs
```diff
@@ -27,34 +28,30 @@ use crate::{
...
     socket: InnerSocket,
-    on: Arc<RwLock<HashMap<Event, Callback<DynAsyncCallback>>>>,
-    on_any: Arc<RwLock<Option<Callback<DynAsyncAnyCallback>>>>,
     outstanding_acks: Arc<RwLock<Vec<Ack>>>,
     // namespace, for multiplexing messages
     nsp: String,
     // Data send in the opening packet (commonly used as for auth)
     auth: Option<serde_json::Value>,
+    builder: Arc<RwLock<ClientBuilder>>,
+    manually_disconnected: Arc<RwLock<bool>>,
```
1c3t3a, line 37: Might as well be an atomic bool?

socketio/src/asynchronous/client/client.rs
```diff
@@ -27,34 +28,30 @@ use crate::{
...
+    server_disconnected: Arc<RwLock<bool>>,
```
1c3t3a, line 38: Same as above?

socketio/src/asynchronous/client/client.rs
```diff
@@ -148,6 +154,10 @@ impl Client {
     /// }
     /// ```
     pub async fn disconnect(&self) -> Result<()> {
+        let mut manually_reconnected = self.manually_disconnected.write().await;
```
1c3t3a, line 157: Variable should be `manually_disconnected` I assume?

socketio/src/asynchronous/client/client.rs
```diff
@@ -375,9 +379,13 @@ impl Client {
                     }
                 }
                 PacketId::Connect => {
+                    let mut server_disconnected = self.server_disconnected.write().await;
+                    *server_disconnected = false;
```
1c3t3a, line 383: For what case is this necessary? Given that we initialize it with false?

1c3t3a (commented):
Thanks a lot for the PR! Really appreciate you working on that :) I left a few comments, generally this looks good.

socketio/src/asynchronous/client/client.rs
rageshkrishna: It gets set to `true` in the `PacketId::Disconnect` case. If we were configured to reconnect and it succeeds, this flag should be reset so it doesn't confuse us when the next disconnection happens (which may not be due to a server disconnect at all).

socketio/src/asynchronous/client/client.rs
rageshkrishna: Makes sense. I've changed both of them

socketio/src/asynchronous/client/builder.rs
rageshkrishna: I agree. I found it odd that the thread was inside the builder instead of the client. Moving it to the client totally makes sense.

rageshkrishna:
@1c3t3a I've incorporated all the suggestions. While adding the test I realized I hadn't made `Client::socket` an Arc. This is necessary because the reconnect thread is working on a clone and we need to be able to share the inner socket instance with all the clones. That's the only big change other than what's in the PR comments.

I made quite a few commits while trying to fix that, but I've squashed everything into just one so it's a clean update. Please take another look and let me know if I've missed something. Thanks!

rageshkrishna:
There's one more problem that I really think should be solved with reconnects. The async client today has no way to update the `auth` field once it's set. This is a problem because, in my case, I'm using short-lived tokens to authenticate the connection. It's very unlikely that this token will be valid when the time comes to reconnect, so we need some way for the user to provide new auth data.

This might be beyond the scope of this PR, though. Maybe I should open another issue for it? Let me know what you think.

socketio/src/asynchronous/client/client.rs
1c3t3a: Makes sense! Thanks for the explanation.

socketio/src/asynchronous/client/client.rs
```diff
@@ -26,49 +35,121 @@ use crate::{
...
+        drop(builder);
+
+        let mut client_clone = self.clone();
+
+        tokio::runtime::Handle::current().spawn(async move {
+            loop {
+                let mut stream = client_clone.as_stream();
+                // Consume the stream until it returns None and the stream is closed.
+                while let Some(item) = stream.next().await {
+                    if let e @ Err(Error::IncompleteResponseFromEngineIo(_)) = item {
```
1c3t3a, line 110: Actually it might make sense to just trace any error here:
```
if let Err(err) = item {
  trace!("Error while polling packet stream: {err}");
}
```

socketio/src/asynchronous/client/client.rs
```diff
@@ -148,11 +233,13 @@ impl Client {
     /// }
     /// ```
     pub async fn disconnect(&self) -> Result<()> {
+        self.manually_disconnected.store(true, Ordering::Relaxed);
```
1c3t3a, line 236: I think this should be `Ordering::Release`? We want to synchronize with every read.

socketio/src/asynchronous/client/client.rs
```diff
@@ -401,11 +484,28 @@ impl Client {
...
+        }
+    }
+
     /// Returns the packet stream for the client.
     pub(crate) fn as_stream<'a>(
         &'a self,
     ) -> Pin<Box<dyn Stream<Item = Result<Packet>> + Send + 'a>> {
-        stream::unfold(self.socket.clone(), |mut socket| async {
+        let socket_clone = {
+            let s = self.socket.try_read().expect("Socket must be readable");
```
1c3t3a, line 504: Mh, this creates a race condition around the lock... which leads to a panic in the background thread... that's not so nice. Any way we can get rid of this and merge this into the stream logic?

1c3t3a (commented):
Looks good!

>  I hadn't made Client::socket an Arc

Yes, that's one necessary step!

> auth field

Maybe better to do this separately! :)

socketio/src/asynchronous/client/client.rs
rageshkrishna: Ahh, ok... something like this?

```rust
        let socket_clone = self.socket.clone();

        stream::unfold(socket_clone, |socket| async {
            let mut socket_read = {
                let s = socket.read().await;
                (*s).clone()
            };
```

rageshkrishna: I've updated the PR with this. I think this avoids any chance of a panic.

socketio/src/asynchronous/client/client.rs
rageshkrishna: Fixed it

socketio/src/asynchronous/client/client.rs
rageshkrishna: Makes sense. Updated.

socketio/src/asynchronous/client/client.rs
rageshkrishna: But this means every iteration clones the socket. Maybe I should just do a `blocking_read` outside?

```rust
        let socket_clone = (*self.socket.blocking_read()).clone();

        stream::unfold(socket_clone, |mut socket| async {
            // wait for the next payload
            let packet: Option<std::result::Result<Packet, Error>> = socket.next().await;
```

This is better, right?

socketio/src/asynchronous/client/client.rs
```diff
@@ -26,49 +35,121 @@ use crate::{
...
+        Ok(())
+    }
+
+    pub(crate) async fn reconnect(&mut self) -> Result<()> {
+        let builder = self.builder.write().await;
+        let socket = builder.inner_create().await?;
+
+        // New inner socket that can be connected
+        let mut client_socket = self.socket.write().await;
+        *client_socket = socket;
```
Colin1860, line 87: Lines 86-88, seem quite counterintuitive can you explain what/why you are doing this?

socketio/src/asynchronous/client/client.rs
```diff
@@ -401,13 +484,31 @@ impl Client {
         Ok(())
     }
 
+    /// Indicates whether the client should try to reconnect
+    pub(crate) async fn should_reconnect(&self) -> bool {
+        let manually_disconnected = self.manually_disconnected.load(Ordering::Acquire);
```
Colin1860, line 489: I think the disconnection handling can be modelled with a tokio channel and a dedicated enum, this would get rid of some of the complexity introduced by the `Arc<AtomicBool>`s which already lead to a small bug which @1c3t3a caught in this PR. 

So essentially you would send over the channel  in line 463 and 236 and matching on the receive here in 489. The enum would essentially have two members now, Manual and Server (e.g). I hope this makes sense to you, but i consider this the more safe and readable solution

Colin1860 (commented):
Hi, @1c3t3a asked me to take a look at this. These are my first thoughts.

socketio/src/asynchronous/client/client.rs
rageshkrishna: Thanks for the input! I'll think about the channels and get back to you in a bit. Can you tell me what bug was found with the use of `AtomicBool`? I don't recall it being mentioned earlier so I'm curious.

socketio/src/asynchronous/client/client.rs
rageshkrishna: I'm replacing `self.socket` with a new connection I've created on line 83. Since `connect()` needs to use it, I have to make sure I'm not holding on to a write lock before I call it on line 90.

socketio/src/asynchronous/client/client.rs
Colin1860: just that you used Ordering::Relaxed instead of Ordering::Release on line 236

rageshkrishna: About channels, I'm not clear on how they would help here. I do agree that replacing the AtomicBools with an enum would make it more readable. So how about something like this:

```rust
pub(crate) async fn should_reconnect(&self) -> bool {
    match self.disconnect_reason.read().await {
      DisconnectReason::Manual => false,
      DisconnectReason::Server => self.builder.read().await.reconnect_on_disconnect
    }
}
```

rageshkrishna: > just that you used Ordering::Relaxed instead of Ordering::Release on line 236

Ahh, right. Yes, that was unintentional but I totally get your point. Accidents are easy with them.

rageshkrishna: Perhaps the enum should also have something like `NetworkError` because that is an expected failure.

Colin1860: Well, I presume you would have to wrap the enum still in an Arc for example, and there would have to be either a Option around your enum or an NoDisconnection (e.g.) variant to model that there was no disconnect. A channel would in that case just be empty. You also don't have to wrap it in any other concurrency primitives like an Arc (or other) that should be handled for you. Also no need to reset the channel, since you only handle the message inside

rageshkrishna: Hmm, channel `Receiver`'s can't be cloned (at least for `mpsc` and `broadcast`). I'm not sure if there's another type of channel that makes sense for this particular case. Well, there's `watch` but that doesn't allow the `Sender` to be cloned. :smile:

It feels like trying to solve this with a channel will not allow us to escape an `Arc`. Do you know if there's another way to do this? My experience with channels isn't extensive, so it is entirely possible that I'm missing something obvious.

rageshkrishna: Even if I wrap a receiver in an Arc, I would need to lock it because it would have to be mutable. I could do all that, but at the end of the day we will still have some synchronization primitive. I don't see any way to escape that because `Client` has to be cloneable. The conclusion I'm reaching is that a channel doesn't really add a lot of value and it seems like a really heavy way to get around having one additional enum value. That being said, I'll defer the final decision to you and @1c3t3a. Let me know how you'd like me to go about it. Thanks!

socketio/src/asynchronous/client/client.rs
1c3t3a: I prefer the latter! :) `blocking_read` seems like the right thing.

socketio/src/asynchronous/client/client.rs
1c3t3a: You can look into `broadcast`: https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html, which allows multiple senders and multiple receivers and doesn't require mutable references for sending and receiving.

1c3t3a: But I agree that the underlying primitives don't get much better. Honestly this is a general problem of this crate since we only create one socket and then use it everywhere, which is not ideal.

Using the `AtomicBool` is equivalent with other code in this crate, however as @Colin1860 rightfully pointed out it is not ideal and easy to shoot ourselves in the foot here.

I would leave the final decision of what to do up to you, or @Colin1860 if he feels strongly about using channels :)

rageshkrishna: Ok, I have replaced both of the `AtomicBool`'s with a `DisconnectReason` enum. I think the code looks much nicer now and is easier to follow. It still needs an `Arc<RwLock>` which is unavoidable.

Even [broadcast::Receiver::recv](https://docs.rs/tokio/1.28.2/tokio/sync/broadcast/struct.Receiver.html#method.recv) needs `&mut self` so I'm not sure it would have made things significantly better in this case.

socketio/src/asynchronous/client/client.rs
rageshkrishna: This is fixed with the latest commit.

socketio/src/asynchronous/client/client.rs
rageshkrishna: I've also added some comments now in the code below to clarify what this is doing.

socketio/src/asynchronous/client/client.rs
1c3t3a: Fair. Lets stick to the atomic bools!

1c3t3a:
Lets run the CI!

1c3t3a:
I fixed the tokio CI issue in main, feel free to rebase :) Then we see the "real" tests of this PR.

rageshkrishna:
Thanks @1c3t3a. Rebased just now. Let's see how it goes! 😄

rageshkrishna:
Well, that was dumb of me. We can't do a `blocking_read` inside `as_stream` because it blocks the runtime. I've changed `as_stream` to be async. I've pushed the update now with all tests passing locally.

I somehow forgot to run the tests after making that change yesterday. Sorry about that!

socketio/src/asynchronous/client/client.rs
```diff
@@ -854,7 +847,7 @@ mod test {
             .await?;
 
         // open packet
-        let mut socket_stream = socket.as_stream();
+        let mut socket_stream = socket.as_stream().await;
         let _ = socket_stream.next().await.unwrap()?;
 
         println!("Here12");
```
1c3t3a, line 853: Just saw this now: Where does that come from? :)

rageshkrishna: It wasn't me. Apparently you added it 12 months ago: https://github.com/1c3t3a/rust-socketio/pull/296/files#diff-dd5d6ff18628a21723847133d6e54ab531454037b62be81d2c42109d333aeab9R609 😆 

I'll remove it.

rageshkrishna:
Is the current failure a flaky test? It's coming from a sync client test, but I'm getting 100% pass locally.

socketio/src/asynchronous/client/client.rs
rageshkrishna: Done.

1c3t3a:
> Is the current failure a flaky test?

I never saw a flaky test in this repo.. that is weird. How do you execute the servers locally? Do you use the devcontainer?

1c3t3a:
Does the server side test have any state? E.g. you first execute the async test and then it doesn't work anymore?
Can you try executing both tests after each other locally?

rageshkrishna:
> I never saw a flaky test in this repo.. that is weird. How do you execute the servers locally? Do you use the devcontainer?

No, I fired up the docker container for the websocket server and ran `cargo test` like it's written in the ci readme.

rageshkrishna:
Ahh, I see what I was missing now... I wasn't passing `--features async` to the test. I'm able to repro this failure on my branch. I'll try to get to the bottom of this.

rageshkrishna:
It was because the new test I added in the async client uses the same `crate::test::socket_io_restart_server()` that the existing test in the sync client uses. Since these tests execute in parallel, they clobber each other's expectations of when the server will connect/disconnect their connections.

My solution to this is adding a new dev-dependency on the [`serial_test`](https://docs.rs/serial_test/latest/serial_test/) crate, and then marking both of these reconnect tests with the `[serial_test(reconnect)]` attribute. `reconnect` is just a key to identify which tests must be executed serially, so this doesn't affect the other tests in any way.

With this, `cargo test --features async` is all green. Please take a look and let me know what you think.

rageshkrishna:
Here's the commit that adds it: https://github.com/1c3t3a/rust-socketio/pull/400/commits/e2eea62f3f8723bcfbbf90c55ebb12487cb5c472

1c3t3a (approved):
LGTM! This was a big one, thanks a lot @rageshkrishna for implementing this, really appreciated :)

PR #405, opened 2024-03-26
Add `on_reconnect` callback on the async client

### LINKED ISSUE #401: Allow async connections to update auth when reconnecting
#291 will add support for reconnections to the async client. Once that lands, we should give the app an opportunity to update its connection settings (address, auth, etc.), because the original settings may not be valid by the time a reconnection is required.

Adding something like an `on_reconnect` callback on the builder would allow us to do this without any breaking changes, like so:

```rust
let client = rust_socketio::asynchronous::ClientBuilder::new(server)
    .on_reconnect(|| {
        async move {
            // Do something async...
            sleep(Duration::from_secs(5)).await;

            // Generate and return new config
            let mut settings = ReconnectSettings::new();
            settings.auth(json!({"token": "foo"}));
            settings
        }
        .boxed()
    })
```

rageshkrishna:
Async clients that are using the reconnect feature can now provide an optional callback that will be invoked before the server is reconnected. You may optionally update the address or auth settings in this callback and they will be applied when the client reconnects.

This is useful for updating auth tokens which may have expired by the time a reconnection is required.

Fixes #401

### DIFF
--- a/socketio/src/asynchronous/client/builder.rs
+++ b/socketio/src/asynchronous/client/builder.rs
@@ -11,8 +11,10 @@ use url::Url;
 use crate::{error::Result, Event, Payload, TransportType};
 
 use super::{
-    callback::{Callback, DynAsyncAnyCallback, DynAsyncCallback},
-    client::Client,
+    callback::{
+        Callback, DynAsyncAnyCallback, DynAsyncCallback, DynAsyncReconnectSettingsCallback,
+    },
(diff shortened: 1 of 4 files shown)


rageshkrishna:
Hi @1c3t3a! Just checking if you had a chance to look at this. Let me know if you need any changes. Thanks.

1c3t3a (approved):
LGTM! This looks great, thanks a lot! :)

PR #421, opened 2024-04-16
[WIP] custom packet serializer implementation

### LINKED ISSUE #406: any support on using msgpack as serializer?

shenjackyuanjie:
yea, here I am again
gogogo!
will fix #406

### DIFF
--- a/engineio/src/header.rs
+++ b/engineio/src/header.rs
@@ -5,7 +5,6 @@ use http::{
     HeaderValue as HttpHeaderValue,
 };
 use std::collections::HashMap;
-use std::convert::TryFrom;
 use std::fmt::{Display, Formatter, Result as FmtResult};
 use std::str::FromStr;
 
--- a/engineio/src/packet.rs
+++ b/engineio/src/packet.rs
@@ -2,7 +2,7 @@ use base64::{engine::general_purpose, Engine as _};
 use bytes::{BufMut, Bytes, BytesMut};
 use serde::{Deserialize, Serialize};
 use std::char;
-use std::fmt::{Display, Formatter, Result as FmtResult, Write};
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult, Write};
 use std::ops::Index;
 
 use crate::error::{Error, Result};
@@ -53,7 +53,6 @@ fn default_encode(packet: Packet) -> Bytes {
     result.freeze()
 }
 
-
 impl PacketSerializer {
     const SEPARATOR: char = '\x1e';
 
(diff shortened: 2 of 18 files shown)


engineio/Cargo.toml
```diff
@@ -14,7 +14,7 @@ license = "MIT"
 all-features = true
 
 [dependencies]
-base64 = "0.21.5"
```
1c3t3a, line 17: Please **don't** do unrelated changes as part of your PRs! Why does this needs to be bumped (also to the next major version, which implies a breaking change).

engineio/Cargo.toml
```diff
@@ -29,7 +29,8 @@ async-trait = "0.1.77"
 async-stream = "0.3.5"
 thiserror = "1.0"
 native-tls = "0.2.11"
-url = "2.4.1"
```
1c3t3a, line 32: Same as above? Why is url bumped?

engineio/src/header.rs
```diff
@@ -5,7 +5,6 @@ use http::{
     HeaderValue as HttpHeaderValue,
 };
 use std::collections::HashMap;
-use std::convert::TryFrom;
```
1c3t3a, line 8: Why is this necessary? Please don't do unrelated changes

engineio/Cargo.toml
```diff
@@ -29,7 +29,8 @@ async-trait = "0.1.77"
...
+url = "2.5.0"
+rmp = "0.8"
```
1c3t3a, line 33: Can we hide this serializer feature behind a feature flag? I don't think we need this everywhere, but rather have it enabled if necessary. For general work with feature flags, you can take a look here: https://doc.rust-lang.org/cargo/reference/features.html.

engineio/src/packet.rs
```diff
@@ -1,13 +1,36 @@
...
 use serde::{Deserialize, Serialize};
 use std::char;
-use std::convert::TryFrom;
-use std::convert::TryInto;
 use std::fmt::{Display, Formatter, Result as FmtResult, Write};
 use std::ops::Index;
 
 use crate::error::{Error, Result};
+
+pub mod normal;
```
1c3t3a, line 11: I think normal is not an ideal name for this. Rather say "standard"?

engineio/src/packet.rs
```diff
@@ -1,13 +1,36 @@
...
+//     MessagePack,
+// }
+
+// impl Default for PacketSerializer {
+//     fn default() -> Self {
+//         PacketSerializer::Normal
+//     }
+// }
+
+pub trait PacketSerializer{
```
1c3t3a, line 29: The trait is much better (in case the commented code presents an alternative), although I am not sure if I am 100% happy with the definition. I am going to think about it :)

1c3t3a (commented):
Hey @shenjackyuanjie, just a quick first pass :)

engineio/Cargo.toml
shenjackyuanjie: I have remove this part of code and left only custom serializer support

shenjackyuanjie:
to be clear, the code is not done, and there is some leftover need to be cleanup

shenjackyuanjie:
bad news, I found my implment should be at socketio packet ……
well, close and anther PR will be send

Issue #422, opened 2024-04-21
when using binary buf data with emit_with_ack function callback is not executed


dontcryme:
If i communicate with simple node.js(socket.io) server with rust's emit_with_ack function(binary buf) then callback function is not executed.
because of rust client's emit_with_ack function send bad ack.

```
[Server Code]

import express from "express";
import http from "http";
import { Server } from "socket.io";
const app = express();
const server = http.createServer(app);
const io = new Server(server);

io.on("connection", (socket) => {
  console.log("Client Connected");

  socket.on("pplong", (data, callback) => {
    console.log("Received Data : ", data);
    callback("true");
  });

});

server.listen("3333", () => {
  console.log("listening on 3333");
});

```

```
[Server Log]

socket.io-parser decoded 61-958["pplong",{"_placeholder":true,"num":0}] as {"type":6,"attachments":1,"nsp":"/","id":958,"data":["pplong",{"_placeholder":true,"num":0}]} +1ms
  engine:ws received "" +1ms
  engine:socket received packet message +1ms
  socket.io:socket got packet {"type":3,"nsp":"/","id":958,"data":["pplong",{"type":"Buffer","data":[0,1,3]}]} +3ms
  socket.io:socket bad ack 958 +1ms
```

The socket.io server handles EVENT and BINARY_EVENT in the following way

```
...
 case PacketType.EVENT:
          case PacketType.BINARY_EVENT:
            this.onevent(packet);
...

```

If modified packet.rs 'new_from_payload' function check logic like below, it will run smoothly without any issues.

```
[Before]

 if id.is_some(){
     PacketId::BinaryAck
 } else {
     PacketId::BinaryEvent
 },

[After]

  if id.is_some() && id.unwrap() == PacketId::Event as i32{
     PacketId::BinaryEvent
  } else {
    PacketId::BinaryAck
  },

```

socket.io-client npm module use same way for sending event and binaryevent.

PR #423, opened 2024-04-21
fix: modify new_from_payload function check logic

dontcryme:
Issue : #422

If i communicate with simple node.js(socket.io) server with rust's emit_with_ack function(binary buf) then callback function is not executed.
because of rust client's emit_with_ack function send bad ack.

so i think that pakcet.rs's new_from_payload function check logic need to be modified like below.

```
[Before]

 if id.is_some(){
     PacketId::BinaryAck
 } else {
     PacketId::BinaryEvent
 },

[After]

  if id.is_some() && id.unwrap() == PacketId::Event as i32{
     PacketId::BinaryEvent
  } else {
    PacketId::BinaryAck
  },
```

### DIFF
--- a/socketio/src/packet.rs
+++ b/socketio/src/packet.rs
@@ -42,10 +42,10 @@ impl Packet {
     ) -> Result<Packet> {
         match payload {
             Payload::Binary(bin_data) => Ok(Packet::new(
-                if id.is_some() {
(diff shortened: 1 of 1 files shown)


dontcryme:
There is an issue, so I will double check.

Issue #427, opened 2024-04-29
Introduce Custom shared data structure


kitgxrl:
This idea was gotten from [serenity-rs/serenity](https://github.com/serenity-rs/serenity), while their implementation is much more nuanced I believe this can be simplified fairly heavily for most use cases.

The idea is to have some data structure that is shared amongst events, possibly accessed via `socket.data()` or something similar. This can indirectly solve #363 and #425

Some example code may be as such, taking example of #363: 
```rs
struct MyData {
    tx: Sender<String>
}

fn main() {
    let (tx, rx) = mpsc::channel::<String>():

    let data = Arc::new(MyData {
        tx
    });

    let callback1 = |payload: Payload, socket: RawClient| {
        let data: Arc<MyData> = socket.data();

        data.tx.send("...".to_string());
    };

    let callback2 = |payload: Payload, socket: RawClient| {
        let data: Arc<MyData> = socket.data();

        data.tx.send("...".to_string());
    };

    let socket = ClientBuilder::new("...")
        .on("event1", callback1)
        .on("event2", callback2)
        .data(data)
        .connect()
        .expect("Connection failed");
}
```
I am absolutely willing to work on this, it is exam week at my college so it might be a week or two before I can get around to it but this implementation would greatly help in cases where shared data is absolutely required.

PR #428, opened 2024-04-29
WIP: Implement shared custom data

kitgxrl:
Implements #427

### DIFF
--- a/socketio/src/client/raw_client.rs
+++ b/socketio/src/client/raw_client.rs
@@ -41,6 +41,7 @@ pub struct RawClient {
     nsp: String,
     // Data send in the opening packet (commonly used as for auth)
     auth: Option<Value>,
+    data: Arc<dyn std::any::Any + Send + Sync>,
 }
 
 impl RawClient {
@@ -54,6 +55,7 @@ impl RawClient {
         on: Arc<Mutex<HashMap<Event, Callback<SocketCallback>>>>,
         on_any: Arc<Mutex<Option<Callback<SocketAnyCallback>>>>,
         auth: Option<Value>,
+        data: Arc<dyn std::any::Any + Send + Sync>,
     ) -> Result<Self> {
         Ok(RawClient {
             socket,
@@ -62,7 +64,19 @@ impl RawClient {
             on_any,
             outstanding_acks: Arc::new(Mutex::new(Vec::new())),
             auth,
+            data,
         })
+    }
+
+    // TODO: Write documentation
+    pub fn data<D: Send + Sync + 'static>(&self) -> Arc<D> {
+        self.try_data()
+            .expect("RawClient::data does not match ClientBuilder::data")
+    }
+
+    // TODO: Write documentation
+    pub fn try_data<D: Send + Sync + 'static>(&self) -> Option<Arc<D>> {
+        Arc::clone(&self.data).downcast().ok()
     }
 
     /// Connects the client to a server. Afterwards the `emit_*` methods can be
--- a/socketio/src/client/builder.rs
+++ b/socketio/src/client/builder.rs
@@ -41,6 +41,7 @@ pub struct ClientBuilder {
     opening_headers: Option<HeaderMap>,
     transport_type: TransportType,
     auth: Option<serde_json::Value>,
+    data: Option<Arc<dyn std::any::Any + Send + Sync>>,
     pub(crate) reconnect: bool,
     pub(crate) reconnect_on_disconnect: bool,
     // None reconnect attempts represent infinity.
@@ -98,7 +99,13 @@ impl ClientBuilder {
             max_reconnect_attempts: None,
             reconnect_delay_min: 1000,
             reconnect_delay_max: 5000,
+            data: None,
         }
+    }
+
+    pub fn data<D: std::any::Any + Send + Sync>(mut self, data: Arc<D>) -> Self {
+        self.data = Some(data);
+        self
     }
 
     /// Sets the target namespace of the client. The namespace should start
@@ -365,6 +372,7 @@ impl ClientBuilder {
             self.on,
             self.on_any,
             self.auth,
+            self.data.unwrap_or(Arc::new(())),
         )?;
         socket.connect()?;
 


kitgxrl:
Will write tests and documentation either when I get home or later tonight!

kitgxrl:
Opening this up for review. I can add tests if they're desired.

socketio/src/asynchronous/client/builder.rs
```diff
@@ -97,9 +99,15 @@ impl ClientBuilder {
             max_reconnect_attempts: None,
             reconnect_delay_min: 1000,
             reconnect_delay_max: 5000,
+            data: None,
         }
     }
 
+    pub fn data<D: std::any::Any + Send + Sync>(mut self, data: Arc<D>) -> Self {
```
1c3t3a, line 106: nit: Maybe name set_data? Didn't find that this is forbidden according to the naming guidelines this crate uses: https://rust-lang.github.io/api-guidelines/naming.html. Only getters omit the prefix.

1c3t3a, line 106: Why already take the Arc? Can't we get a fat pointer? Or does that play out really badly with lifetimes?

1c3t3a, line 106: Please add documentation for the method. IMO this is the perfect place for explaining what data actually is.

socketio/src/asynchronous/client/client.rs
```diff
@@ -87,11 +88,25 @@ impl Client {
...
+            data: builder.data.clone().unwrap_or(Arc::new(())),
             builder: Arc::new(RwLock::new(builder)),
             disconnect_reason: Arc::new(RwLock::new(DisconnectReason::default())),
         })
     }
 
+    /// Fetches data given by [`ClientBuilder::data`]
+    pub fn data<D: Send + Sync + 'static>(&self) -> Arc<D> {
+        self.try_data()
+            .expect("Client::data does not match ClientBuilder::data")
```
1c3t3a, line 100: Mhmmm... I don't really like that we need a runtime check here... Wdyt, would it be possible to make this a generic parameter of both the Client and the builder? If this turns out messy, feel free to omit.

socketio/src/asynchronous/client/client.rs
```diff
@@ -87,11 +88,25 @@ impl Client {
...
+    }
+
+    /// Attempts to fetch data given by [`ClientBuilder::data`]
+    ///
+    /// None is returned if data was not given or data does not match [`ClientBuilder::data`]
+    pub fn try_data<D: Send + Sync + 'static>(&self) -> Option<Arc<D>> {
```
1c3t3a, line 106: Mhm, I don't really like the naming here. Usually a `try_` prefix implies a Result, which we don't have here. How about we name this method `data` (returning an option) and the other one `data_unchecked` which panics? Or we actually don't provide a panic method (because that's usually a bad idea anyways, no library should panic), and just make the upper one return a Result?

socketio/src/asynchronous/client/client.rs
```diff
@@ -87,11 +88,25 @@ impl Client {
...
             auth: builder.auth.clone(),
+            data: builder.data.clone().unwrap_or(Arc::new(())),
             builder: Arc::new(RwLock::new(builder)),
             disconnect_reason: Arc::new(RwLock::new(DisconnectReason::default())),
         })
     }
 
+    /// Fetches data given by [`ClientBuilder::data`]
+    pub fn data<D: Send + Sync + 'static>(&self) -> Arc<D> {
+        self.try_data()
```
1c3t3a, line 99: Panicing in library code is always a risky thing. We rather want controllable errors (`Result`s).

socketio/src/client/raw_client.rs
```diff
@@ -62,9 +65,23 @@ impl RawClient {
...
+            .expect("RawClient::data does not match ClientBuilder::data")
+    }
+
+    /// Attempts to fetch data given by [`ClientBuilder::data`]
+    ///
+    /// None is returned if data was not given or data does not match [`ClientBuilder::data`]
+    pub fn try_data<D: Send + Sync + 'static>(&self) -> Option<Arc<D>> {
+        Arc::clone(&self.data).downcast().ok()
+    }
+
```
1c3t3a, line 84: Same comments as above!

1c3t3a (commented):
Hey @kitgxrl, first of all thanks a lot for the PR! I really like the idea. I added some comments to the code on a first pass.

Also I would like to discuss the naming: I think `data` is a bit misleading. Reading this I would think I get internal connection data of the protocol/client. Instead it'd be better to have a name that directly implies that all control of that "data" is with the user. So e.g. `shared_data`, `custom_data` or sth along these lines (I didn't think too long about this). What's your opinion?

kitgxrl:
Hey @1c3t3a thanks for the review! I'll take a look at the comments made sometime Thursday or Friday due to exams over the next two days (I apologize for the late response).

Both `custom_data` and `shared_data` sound good to me, I didn't think about how `data` could be misleading, thanks for pointing that out. If I think of something better I'll be sure to bring it up!

1c3t3a:
No worries, there is absolutely now need to apologize! I am super happy you work on this in your free time :) 

Good luck with the exams! And in that case I'd personally prefer `custom_data`, but feel free to come up with something different.

socketio/src/asynchronous/client/client.rs
kitgxrl: Generic parameter was my main concern, sounds like it'd get messy and fast.

socketio/src/asynchronous/client/builder.rs
kitgxrl: set_data sounds good to me.

kitgxrl: Yeah! Sure, I mainly forgot to add documentation here honestly.

socketio/src/asynchronous/client/client.rs
kitgxrl: `data_unchecked` doesn't really sit right with me. Maybe just `try_data` returning the result is better suited.

kitgxrl: Or rather just having a `custom_data` function that returns the result might be better, not `try_data`

kitgxrl:
@1c3t3a  Took a look at most of what you said, let me know if I missed anything big you'd like to go over. Only thing I believe I haven't looked into yet is using generics on Client and ClientBuilder and accepting a fat pointer on `set_data`, I'll try and look over this in the next day or two when I have a chance. Apologies for going over this so late haha.

Morb0:
Any chance?

PR #435, opened 2024-05-14
Implement shared custom data (re-worked) based on the previous work from kitgxrl

thebashpotato:
Greetings, just wanted to say this is a great library and you've done fantastic work.
Just wanted to throw my hat in the ring to solve the shared data issue.

I read the the discussion on kitgxrl's PR and tried to solve the comments you mentioned.
I decided to go with the naming convention of 'transmitter' as that's really what's going on.
I stayed with Arc, as a Fat pointer caused way too many lifetime issues, although I did give it the good ol' college try, in the end an Arc isn't that much overhead.

I also fixed all the old clippy warnings that was causing the `make clippy` command to complain. This included removing the two client/client.rs modules as its a module naming repetition warning. I just put the code in the corresponding client/mod.rs module instead.

I wrote two comprehensive example clients for both async and raw clients, added proper documentation.

I also added an error, and went with `try_transmitter` which returns a Result, to ensure the library can't panic.

If there is anything else you would like done, please let me know.


kitgxrl:
@thebashpotato I'm a bit confused about the naming convention of `Transmitter`, could you explain a bit more? Sorry if this is a bit of a silly question. Seems like this is assuming that this is some sort of channel which isn't necessarily true. For my use case for example, this is used more as a way of shared state than just having a channel, it just happen to be one of the issues some had that implementing custom data would fix.

socketio/examples/async_transmitter.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+impl TransmitterClient {
+    async fn connect(url: &str) -> Result<Self, SocketIOError> {
+        let (sender, receiver) = mpsc::channel::<String>();
+
+        let client = SocketIOClientBuilder::new(url)
+            .namespace("/admin")
+            .on("test", |payload: Payload, socket: SocketIOClient| {
+                async move {
```
1c3t3a, line 30: Can't we put the `async move` after the closue, e.g.:
`.on("test", |payload: Payload, socket: SocketIOClient| async move {`? That way we have one nesting less.

socketio/examples/async_transmitter.rs
```diff
@@ -0,0 +1,116 @@
...
+                                                            println!(
+                                                                "Data transmitted successfully"
+                                                            )
+                                                        },
+                                                    );
+                                            },
+                                        );
+                                }
+                            }
+                        }
```
1c3t3a, line 53: I am not a huge fan of the nesting tbh :D Could we either (or both):
1. Seperate this out into it's own async function?
2. Since we're only interested in `Payload::Text`, maybe we can shrink this further by having `if let Payload::Text(values) = payload {`

socketio/src/asynchronous/client/async_client.rs
```diff
@@ -87,11 +88,71 @@ impl Client {
...
+    ///                         );
+    ///                     }
+    ///                 }
+    ///             }
+    ///             Payload::Binary(bin_data) => println!("{:#?}", bin_data),
+    ///             #[allow(deprecated)]
+    ///             Payload::String(str) => println!("Received: {}", str),
+    ///         }
+    ///     }
+    ///     .boxed()
```
1c3t3a, line 146: Same feedback as in the example :)

socketio/src/asynchronous/client/builder.rs
```diff
@@ -97,9 +97,43 @@ impl ClientBuilder {
...
+            transmitter: None,
         }
     }
 
+    /// Sets the data transmission object, ideally the standard libraries
+    /// multi-producer single consumer [`std::sync::mpsc::Sender`] should be used.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
```
1c3t3a, line 109: Why no_run? I assume it disables doc tests? Does it compile them?

socketio/src/client/client.rs
```diff
@@ -1,478 +0,0 @@
-use std::{
```
1c3t3a, line 1: Same comment as for the other client, I'd like to keep it, for now!

socketio/src/client/raw_client.rs
```diff
@@ -62,9 +64,58 @@ impl RawClient {
...
+    ///                        },
+    ///                    );
+    ///                }
+    ///            }
+    ///        }
+    ///        Payload::Binary(bin_data) => println!("{:#?}", bin_data),
+    ///        #[allow(deprecated)]
+    ///        Payload::String(str) => println!("Received: {}", str),
+    ///    }
+    /// };
```
1c3t3a, line 110: Same feedback as for the rest of similar code.

1c3t3a (commented):
Hey @thebashpotato thanks a lot for PR!! I really like the channel design here and I agree with the name `transmitter`! I did a first pass over this code and gave some feedback, also I enabled CI runs.

@kitgxrl: Since we now have two PRs with the same content, let me think about how we proceed.

kitgxrl:
@1c3t3a Sure! Although, I'm still a bit confused with `transmitter` as the naming here, honestly.

engineio/src/client/client.rs
```diff
@@ -1,662 +0,0 @@
-use super::super::socket::Socket as InnerSocket;
```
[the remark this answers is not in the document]
thebashpotato: Yes, it was to fix the lint, I figured since the lint was enabled that it just hadn't got around to being fixed yet, I can put the code back, and disable the lint.

socketio/examples/async_transmitter.rs
thebashpotato: For sure, no problem

kitgxrl:
@thebashpotato Sure, however you're able to store arbitrary data here, which was my original intention. `transmitter` I feel is a bit misleading of how it can actually be used. `custom_data` or something similar still feels a bit better as it's made clear arbitrary data is able to be stored, not just a channel sender.

thebashpotato:
@kitgxrl Fair enough, I leave the decision up to @1c3t3a as the repo owner. I'll name it whatever, in the mean time I'll continue fixing the other issues.

shenjackyuanjie:
HI!
I like the idea of this PR also
at the same time, I'm trying to implement [Custom parser](https://socket.io/docs/v4/custom-parser/) for socketio Packet
so, there may be some part of code need to be changed later for supporting parser

socketio/src/packet.rs
```diff
@@ -216,7 +215,7 @@ impl TryFrom<&Bytes> for Packet {
     /// this member. This is done because the attachment is usually
     /// send in another packet.
     fn try_from(payload: &Bytes) -> Result<Packet> {
-        let mut payload = str_from_utf8(&payload).map_err(Error::InvalidUtf8)?;
+        let mut payload = str_from_utf8(payload).map_err(Error::InvalidUtf8)?;
         let mut packet = Packet::default();
```
shenjackyuanjie, line 220: This part will be replaced with some kind of custom parser

1c3t3a: Thanks for the heads up!

PR #449, opened 2024-07-16
feat(socketio): notify client of transport completion

### LINKED ISSUE #448: Async client is not notified of (unexpected) websocket transport close
## Problem

Using the async client without reconnect, when the transport websocket closes properly but without an engineio message indicating the close, the client doesn't fire an event and as such the user of the client doesn't notice the closure.

This has the problem that we can't easily react to the closure and keep thinking that we are connected.

## Steps to reproduce

1. Run a server like this:

    ```javascript
    const server = require('http').createServer();
    const io = require('socket.io')(server);

    console.log('Started');
    var callback = client => {
        console.log('Connected!');

        // Close underlying websocket connection
        client.client.conn.close();
    };
    io.on('connection', callback);
    // the socket.io client runs on port 4201
    server.listen(4200);
    ```

1. Connect with a Rust client like the following. Make sure to set `reconnect` to false. To get some more logging, you can run `export RUST_LOG="info,tungstenite::protocol=trace,tungstenite::protocol::frame::frame=debug,rust_socketio=trace,rust_engineio=trace"` first.

    ```rust
    use futures_util::FutureExt;
    use log::{error, info};
    use rust_socketio::{asynchronous::ClientBuilder, Event, TransportType,};
    use tokio::signal::{self,unix::{signal, SignalKind}};

    #[tokio::main]
    async fn main() {

        info!("Starting");

            .reconnect(false)
            .on(Event::Connect, |_payload, _client| {{ async move { info!("Connected!"); }.boxed() }})
            .on(Event::Close, |_payload, _client| {{ async move { info!("Connection closed!"); }.boxed() }})
            .on(Event::Error, |err, _| { async move { error!("Error: {:#?}", err) }.boxed() })
            .connect()
            .await
            .expect("Connection failed");

        tokio::select! {
            _ = signal::ctrl_c() => { info!("Received SIGINT, shutting down."); },
            _ = sigterm.recv() => { info!("Received SIGTERM, shutting down."); },
        }

    }
    ```

1. Notice that the tungstenite websocket closes by receiving an sending an `opecode: CLOSE` frame, but the client isn't notified.
    
## Reason

When the async client connects to the server, it spawns a new thread to handle the arriving messages asynchronously, which is immediately detached with no chance of awaiting its completion.

This may happen if the underlying tungstenite websocket shuts down ('Received close frame: None') but there are no engineio/socketio close frames. Hence, since the stream terminates, the message handling task stops without a Close or Error event being fired.

https://github.com/1c3t3a/rust-socketio/blob/64d2a133560aa0b31b2263a820645f982e7276b3/socketio/src/asynchronous/client/client.rs#L151-L159

This was kind of introduced by me in a9a81e9 when I added handling for the stream closure to avoid the panic. As it says in the commit message *However, the program which uses the socket.io client currently isn't notified of the unexpected closure.*, so time to clean that up :)

## Potential Solution

Looking at the official JavaScript [`disconnect` specification](https://socket.io/docs/v4/client-socket-instance/#disconnect) there are reasons specified for the disconnection which include a `transport close`.

Thus, once we disconnect for an unknown reason (e.g. the transport closed), we could fire a `CLOSE` event to signal the closure to the user.

Alternatively, we could also fire this event only after the loop terminates, before the thread itself terminates.

See #449 for an implementation of that and feel free to discuss the proposed changes.

sirkrypt0:
When the async client connects to the server, it spawns a new thread to handle the arriving messages asynchronously, which is immediately detached with no chance of awaiting its completion.

For long-running programs (e.g. a client program that never really disconnects from the server) this can be problematic, as unexpected stream completions would go unnoticed. This may happen if the underlying tungstenite websocket shuts down ('Received close frame: None') but there are no engineio/socketio close frames. Hence, since the stream terminates, the message handling task stops without a Close or Error event being fired.

Thus, we now fire an additional Event::Close when the stream terminates, to signal the (potentially unexpected) close to the user.

Closes #448 

Once we settle on a final solution, I can also write a test for that.

### DIFF
--- a/ci/socket-io.js
+++ b/ci/socket-io.js
@@ -36,6 +36,14 @@ var callback = client => {
             ack(Buffer.from([1, 2, 3]));
         }
     });
+
+    // This event allows the test framework to arbitrarily close the underlying connection
+    client.on('close_transport', data => {
+        console.log(['close_transport', 'Request to close transport received'])
+        // Close underlying websocket connection
+        client.client.conn.close();
+    })
+
     client.emit('Hello from the message event!');
     client.emit('test', 'Hello from the test event!');
     client.emit(Buffer.from([4, 5, 6]));
--- a/socketio/src/asynchronous/client/client.rs
+++ b/socketio/src/asynchronous/client/client.rs
@@ -2,7 +2,7 @@ use std::{ops::DerefMut, pin::Pin, sync::Arc};
 
 use backoff::{backoff::Backoff, ExponentialBackoffBuilder};
 use futures_util::{future::BoxFuture, stream, Stream, StreamExt};
-use log::trace;
+use log::{error, trace};
 use rand::{thread_rng, Rng};
 use serde_json::Value;
 use tokio::{
@@ -162,7 +162,21 @@ impl Client {
                 drop(stream);
 
                 let should_reconnect = match *(client_clone.disconnect_reason.read().await) {
-                    DisconnectReason::Unknown => reconnect,
+                    DisconnectReason::Unknown => {
+                        // If we disconnected for an unknown reason, the client might not have noticed
+                        // the closure yet. Hence, fire a transport close event to notify it.
+                        // We don't need to do that in the other cases, since proper server close
+                        // and manual client close are handled explicitly.
+                        if let Some(err) = client_clone
+                            .callback(&Event::Close, "transport close")
+                            .await
+                            .err()
+                        {
+                            error!("Error while notifying client of transport close: {err}")
+                        }
+
+                        reconnect
+                    }
                     DisconnectReason::Manual => false,
                     DisconnectReason::Server => reconnect_on_disconnect,
                 };
@@ -590,6 +604,7 @@ mod test {
         },
         error::Result,
         packet::{Packet, PacketId},
+        Event,
         Payload, TransportType,
     };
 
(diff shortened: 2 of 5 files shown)


MnlPhlp:
Are there any remaining problems with this?
This looks like it solves exactly what I need.

ctrlaltf24 (approved):

socketio/src/asynchronous/client/client.rs
```diff
@@ -162,7 +162,21 @@ impl Client {
...
 
                 let should_reconnect = match *(client_clone.disconnect_reason.read().await) {
-                    DisconnectReason::Unknown => reconnect,
+                    DisconnectReason::Unknown => {
+                        // If we disconnected for an unknown reason, the client might not have noticed
+                        // the closure yet. Hence, fire a transport close event to notify it.
+                        // We don't need to do that in the other cases, since proper server close
+                        // and manual client close are handled explicitly.
+                        if let Some(err) = client_clone
+                            .callback(&Event::Close, "transport close")
```
ctrlaltf24, line 171: nit: constant for this message

ctrlaltf24 (commented):
suggestion: would be nice to have a test for this before merging

socketio/src/asynchronous/client/client.rs
sirkrypt0: I added a new `CloseReason` enum and used the appropriate values wherever the `Event::Close` was emitted. This should make it more consistent and easier for clients to distinguish the close reasons.

ci/socket-io.js
```diff
@@ -36,6 +36,13 @@ var callback = client => {
             ack(Buffer.from([1, 2, 3]));
         }
     });
+
+    client.on('transport_close', data => {
+        console.log(['transport_close', 'Transport close received'])
+        // Close underlying websocket connection
+        client.client.conn.close();
+    })
```
ctrlaltf24, line 44: nit: switch to verb then subject as we're trying to tell the server to do something, and add a comment saying what it does/why it's here. This is NOT the feature, merely used in testing.
```suggestion
    // This event allows the test framework to arbitrarily close the underlying connection
    client.on('close_transport', data => {
        console.log(['transport_close', 'Transport close received'])
        // Close underlying websocket connection
        client.client.conn.close();
    })
```

socketio/src/asynchronous/client/client.rs
```diff
@@ -926,6 +941,50 @@ mod test {
...
+            })
+            .connect()
+            .await?;
+
+        // Wait until socket is connected
+        let connect_timeout = timeout(Duration::from_secs(1), notify.notified()).await;
+        assert!(connect_timeout.is_ok());
+
+        // Instruct server to close transport
+        let result = socket.emit("transport_close", Payload::from("")).await;
```
ctrlaltf24, line 973: nit: switch to verb/subject as we're telling the server to do something.
```suggestion
        let result = socket.emit("close_transport", Payload::from("")).await;
```

socketio/src/event.rs
```diff
@@ -57,3 +57,32 @@ impl Display for Event {
...
+    IOServerDisconnect,
+    IOClientDisconnect,
+    TransportClose,
+}
+
+impl CloseReason {
+    pub fn as_str(&self) -> &str {
+        match self {
+            CloseReason::IOServerDisconnect => "io server disconnect",
+            CloseReason::IOClientDisconnect => "io client disconnect",
```
ctrlaltf24, line 72:
```suggestion
            // Inspired by https://github.com/socketio/socket.io/blob/d0fc72042068e7eaef448941add617f05e1ec236/packages/socket.io-client/lib/socket.ts#L865
            CloseReason::IOServerDisconnect => "io server disconnect",
            // Inspired by https://github.com/socketio/socket.io/blob/d0fc72042068e7eaef448941add617f05e1ec236/packages/socket.io-client/lib/socket.ts#L911
            CloseReason::IOClientDisconnect => "io client disconnect",
```

ctrlaltf24 (approved):
LGTM.

@1c3t3a this ready to merge?

ctrlaltf24 (approved):
Still waiting on @1c3t3a

1c3t3a (approved):
LTGM! Sorry for the delay!

PR #463, opened 2024-09-16
Add ability to handle events that expect an ack

### LINKED ISSUE #461: Receiving a message from the server that requires an ack
I want to send from the socket io server an `emit_with_ack` event. But I don't see how I can ack from the client

```rs
async fn handler(payload: Payload, socket: Client) {
    let values = match payload {
        Payload::Text(values) => values,
        Payload::Binary(_) => panic!("unexpected bytes"),
        #[allow(deprecated)]
        Payload::String(_) => panic!("Payload::String panicked"),
    };

    tracing::info!("got values: {values:?}");
    // how do I ack here?
}

async fn run(config: &Config, hostname: &Hostname) -> anyhow::Result<()> {
    let socket = ClientBuilder::new(format!("{}?h={}", config.backend_domain, hostname))
        .namespace(ws::NS)
        .on(ws::COMMAND, |payload, socket| {
            handler(payload, socket).boxed()
        })
        .on("error", |err, _| {
            async move { tracing::error!(error = ?err, "socket io error") }.boxed()
        })
        .connect()
        .await?;

    std::future::pending::<()>().await;
    socket.disconnect().await?;
    drop(socket);
    Ok(())
}
```

I've tested and I do get a value in here but I can't ack. Is this just not implemented yet or am I missing something?

mendess:
This adds three new methods to the API surface

- `[Raw]Client::ack(AckId, D)`
- `ClientBuilder::on_with_ack(Event, callback)`
- `ClientBuilder::on_any_with_ack(Event, callback)`

using these it's possible to receive messages that require ack and acknowledge them

Fixes #461

### DIFF
--- a/socketio/src/client/builder.rs
+++ b/socketio/src/client/builder.rs
@@ -180,6 +180,41 @@ impl ClientBuilder {
         self
     }
 
+    /// Registers a new callback for a certain [`crate::event::Event`] that expects the client to
+    /// ack. The event could either be one of the common events like `message`, `error`, `open`,
+    /// `close` or a custom event defined by a string, e.g. `onPayment` or `foo`.
+    ///
+    /// # Example
+    /// ```rust
+    /// use rust_socketio::{ClientBuilder, Payload};
+    ///
+    /// let socket = ClientBuilder::new("http://localhost:4200/")
+    ///     .namespace("/admin")
+    ///     .on_with_ack("test", |payload: Payload, client, ack_id| {
+    ///            match payload {
+    ///                Payload::Text(values) => println!("Received: {:#?}", values),
+    ///                Payload::Binary(bin_data) => println!("Received bytes: {:#?}", bin_data),
+    ///                // This payload type is deprecated, use Payload::Text instead
+    ///                Payload::String(str) => println!("Received: {}", str),
+    ///            }
+    ///            client.ack(ack_id, "received");
+    ///     })
+    ///     .on("error", |err, _| eprintln!("Error: {:#?}", err))
+    ///     .connect();
+    ///
+    /// ```
+    // While present implementation doesn't require mut, it's reasonable to require mutability.
+    #[allow(unused_mut)]
+    pub fn on_with_ack<T: Into<Event>, F>(mut self, event: T, callback: F) -> Self
+    where
+        F: FnMut(Payload, RawClient, i32) + 'static + Send,
+    {
+        let callback = Callback::<SocketCallback>::new_with_ack(callback);
+        // SAFETY: Lock is held for such amount of time no code paths lead to a panic while lock is held
+        self.on.lock().unwrap().insert(event.into(), callback);
+        self
+    }
+
     /// Registers a Callback for all [`crate::event::Event::Custom`] and [`crate::event::Event::Message`].
     ///
     /// # Example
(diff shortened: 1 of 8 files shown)


mendess:
I didn't add any tests because I didn't have the time to do it yet. But I've tested locally with my usecase and it seems to work

socketio/src/client/builder.rs
```diff
@@ -201,6 +236,36 @@ impl ClientBuilder {
...
+    /// Registers a Callback for all [`crate::event::Event::Custom`] and
+    /// [`crate::event::Event::Message`] that expects the client to ack.
+    ///
+    /// # Example
+    /// ```rust
+    /// use rust_socketio::{ClientBuilder, Payload};
+    ///
+    /// let client = ClientBuilder::new("http://localhost:4200/")
+    ///     .namespace("/admin")
+    ///     .on_any(|event, payload, client, ack_id| {
```
ctrlaltf24, line 255:
```suggestion
    ///     .on_any_with_ack(|event, payload, client, ack_id| {
```

socketio/src/packet.rs
```diff
@@ -25,7 +25,7 @@ pub struct Packet {
     pub packet_type: PacketId,
     pub nsp: String,
     pub data: Option<String>,
-    pub id: Option<i32>,
+    pub id: Option<AckId>,
```
ctrlaltf24, line 28: While I'm generally a fan of wrapper types, it's not worth the breaking API change.

ctrlaltf24 (commented):
suggestion: rename added internal functions to ..._with_awk (rather than having _no_awk variants)

suggestion: add tests so the feature continues to be tested and working

after that it looks good

socketio/src/packet.rs
mendess: the reason I put this in is just so you can't make a mistake when calling `ack` on the client. Since you're not supposed to care what the ack id is I think this is a good change and it's better to make this change sooner rather than later. If you still disagree I can drop it

mendess:
> suggestion: add tests so the feature continues to be tested and working

I could not figure out how to run the `test-all` recipe of the makefile, hence I didn't write any tests. I think docs could be improved on this in a later PR. For now can you tell me how I can run the tests?

socketio/src/packet.rs
ctrlaltf24: I agree it's a reasonable change, but it would need to be made (and merged) separately.

ctrlaltf24:
> > suggestion: add tests so the feature continues to be tested and working
> 
> I could not figure out how to run the `test-all` recipe of the makefile, hence I didn't write any tests. I think docs could be improved on this in a later PR. For now can you tell me how I can run the tests?

agreed, that target is a little lacking. You need to start up the test containers beforehand as seen in the ci: https://github.com/1c3t3a/rust-socketio/blob/2ef32ecbe053d100e350c4d77a71dd1cab19471e/.github/workflows/test.yml#L36C11-L37C133, speaking of which I'll turn that on for you (it'll run on every push if you can't get it running locally)

socketio/src/packet.rs
1c3t3a: I also agree that this is a reasonable change, but please factor It out in separate commit / PR. This breaks existing users that depend on an `i32` in this place and we cannot just ignore that. That being said, thanks for spotting this appropriately :)

1c3t3a (commented):
First of all, sorry for the late response.

Thanks a lot for working on this, it generally LGTM!

Regarding the tests: It'd be great to have them and I improved the `make test-all` target in #477.

Would be amazing if you could add the tests and split the AckId change out, then we're ready to merge this!

mendess:
Wtf github? I didn't close this :thinking: maybe it got confused when I was syncing my fork

mendess:
I have removed the ack id commit and will make a separate PR for it later. I've run the `test-all` tests locally and it all passes. I also made a small modification to the makefile so that I could run the tests multiple times

ewancg:
@1c3t3a I saw this in passing & it looks like a review would be appreciated

Issue #374, opened 2023-11-06
Manual disconnect doesn't block reconnecting


jexjang:
# Unable to Disconnect Manually
There seems to be no direct method to manually disconnect. When invoking the `disconnect` method on the client struct, it does trigger an actual disconnection.
However, the reconnection mechanism remains active, which results in the client attempting to reconnect immediately after disconnection.

This issue can be circumvented by setting the reconnect attribute to false in the ClientBuilder, but this solution is not viable for my use case. I need the client to automatically reconnect in the event of an accidental disconnection, **not** a manual one.

Is there a way to manually disconnect the client such that it doesn't attempt to reconnect, while still retaining the automatic reconnection capability in cases of unintentional disconnections?

jexjang:
I think it would be nice if the thread that controls the reconnection mechanism dies.
I can make a PR if you want.

jexjang:
Even if the Client is dropped after the disconnection, the reconnection mechanism stays alive.

PR #466, opened 2024-09-23
fix: manual disconnect not preventing reconnection

SenseiHiraku:
This pull request fixes the issue where calling disconnect() on the client does not prevent the client from automatically reconnecting, as mentioned in #374 

I noticed that the asynchronous client behaves differently since the `DisconnectReason` enum was introduced via #400. I have reimplemented this functionality for the synchronous client. During the process, I encountered issues such as "invoking disconnect after the client lost connection to the server still tries to reconnect," but I have fixed and tested every edge case I could think of, and all should be properly handled now.

### DIFF
--- a/socketio/src/client/client.rs
+++ b/socketio/src/client/client.rs
@@ -3,7 +3,7 @@ use std::{
     time::Duration,
 };
 
-use super::{ClientBuilder, RawClient};
+use super::{raw_client::DisconnectReason, ClientBuilder, RawClient};
 use crate::{
     error::Result,
     packet::{Packet, PacketId},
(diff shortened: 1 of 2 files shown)


1c3t3a (changes requested):
Sorry for letting you wait so long! 

This generally LGTM :) It would just be great if you could add a test for this behavior, so that we notice any regressions in the future. Also please rebase on the latest main, as there is not a merge conflict in one of the files.

Thanks a lot for your contribution and working on this. It is very much appreciated.

vanics:
Why this hasn't been merged? Was this fixed?

This seems to be a critical bug.
Is this library being maintained?

PR #467, opened 2024-09-26
fix: deadlock on recursive emit_with_ack calls

bastilimbach:
Hi @1c3t3a, I came across an issue while trying to call `emit_with_ack` recursively like so:
```rust
socket
        .emit_with_ack(
            "event_with_ack",
            json!({"event_with_ack": "first call"}),
            Duration::from_secs(20000),
            move |data, s| {
                println!("Ack 1 received: {:#?}", data);

                // This works
                s.emit("event", json!({"event": "second call"}))
                    .expect("Server unreachable");

                // This doesn't work due to deadlock on "outstanding_acks" in emit_with_ack(). The callback is never executed.
                s.emit_with_ack(
                    "event_with_ack",
                    json!({"event_with_ack": "second call"}),
                    Duration::from_secs(20000),
                    |data, _| {
                        println!("Ack 2 received: {:#?}", data);
                    },
                )
                .expect("Server unreachable");
            },
        )
        .expect("Server unreachable");
```

I noticed that **the second callback is never executed**, and the program just waits indefinitely. After some quick debugging, I drilled down the problem to a deadlock of `self.outstanding_acks` in the RawClient.

# Steps to reproduce
I created a minimal example which can be used to reproduce the issue: https://github.com/bastilimbach/rust-socketio-deadlock

The steps are written in the [README](https://github.com/bastilimbach/rust-socketio-deadlock#quick-start).

# Remarks
I'm quite new to Rust and would appreciate any requests to changes, etc.
Also, I wasn't sure how to write tests for this. I could create a test which calls `emit_with_ack` recursively and call it a day but I'm not sure if this is really the way to go here. Looking forward to suggestions!

### DIFF
--- a/socketio/src/client/raw_client.rs
+++ b/socketio/src/client/raw_client.rs
@@ -277,11 +277,18 @@ impl RawClient {
             return Ok(());
         };
 
-        self.outstanding_acks.lock()?.retain_mut(|ack| {
-            if ack.id != id {
-                return true;
-            }
+        let outstanding_ack = {
+            let mut outstanding_acks = self.outstanding_acks.lock()?;
+            outstanding_acks
+                .iter()
+                .position(|ack| ack.id == id)
+                .map(|pos| outstanding_acks.remove(pos))
+        };
 
+        // If we found a matching ack, call its callback otherwise ignore it.
+        // The official implementation just removes the ack id on timeout:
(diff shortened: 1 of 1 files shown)


1c3t3a (approved):
LGTM! Well spotted, thank you a lot!

